PR #31, opened 2022-04-09
Add syntax highlight for commands, and fix a bug

Sunt-ing:
1. Add syntax highlight
2. In RisingWave, we use ```shell, so I think here we can also use this format.
3. Fix a bug related to the link? Not sure whether the bug is valid.


cloudcarver (approved):

Sunt-ing:
Anyway, merge it first. If there is any problem with this pull request, we can revert it or modify it then.

PR #38, opened 2022-04-09
disable indexing style

cloudcarver:


cloudcarver:
The style is broken in Safari. Disable the customized indexing style for now

PR #53, opened 2022-04-18
refactor: change link for edit button to the real edit page

Sunt-ing:
IMO usually edit link should be linked to the edit page. Not sure about this.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -31,13 +31,11 @@ const config = {
               path: '/latest',
             }
           },
-          // Please change this to your repo.
-          editUrl: 'https://github.com/singularity-data/risingwave-docs/blob/main/',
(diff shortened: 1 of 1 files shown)


cloudcarver:
cc. @CharlieSYH

Sunt-ing:
We find that some open-source projects choose to remain this way: "\*/blob/\*". Some we decide to use this way also. Example: https://github.com/apache/shardingsphere/edit/master/docs/document/content/overview/_index.en.md

Sunt-ing:
We find that some open-source projects choose to remain this way: "*/blob/*". Some we decide to use this way also.

CharlieSYH (approved):
To conclude, the "Edit this page" button still goes to the 'blob' not 'edit' page because it's easier to preview the file content.

PR #71, opened 2022-05-17
feat: (sticky-table-header): fit dark mode and light mode

### LINKED ISSUE #61: Table header should stay on top of the table
When scrollling down the page, the user may find it difficult to read the table content because the table header is out of sight:
image (2)

The solution brought up by @Ad-Bean might be suitable for this case: https://codepen.io/tjvantoll/pen/nxbaMe

Ad-Bean:
~~use position: sticky and media query to stick the table header when user's screen width is less than 1000px~~

it supports all devices now, when the table's height is more than 50vh, it will show the scroll bar and sticky table header

~~known bug: when the table header is sticked and then click the nav bar, you'll get:~~

Close #61

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -123,27 +123,62 @@ h2 {
   border-bottom: 1px solid #dbdbdb;
 }
 
-@media (max-width: 1000px) {
-  table {
-    position: relative;
-  }
+table {
+  position: relative;
+}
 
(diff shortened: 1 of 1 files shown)


cloudcarver:
Seems this feature is not for desktop?

CharlieSYH (approved):
Looks great! Thanks a lot @Ad-Bean.

PR #90, opened 2022-05-26
Add document for EXPLAIN command

fuyufjh:
As title.

### DIFF
--- a/docs/Commands.md
+++ b/docs/Commands.md
@@ -30,8 +30,7 @@ RisingWave supports the following SQL commands.
 
 ## Notes about EXPLAIN
 
-`EXPLAIN` prints a execution plan of query to help you understand how it run internally. It's useful to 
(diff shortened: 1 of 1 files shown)


fuyufjh:
By the way, Perhaps we may split the `Commands.md` into multiple files:

- DML: Select, Update, Delete...
- DDL: CREATE TABLE, Drop table ...
- DAL: SHOW...

CharlieSYH (approved):
Corrected minor mistakes.

hengm3467:
> By the way, Perhaps we may split the `Commands.md` into multiple files:
> 
> * DML: Select, Update, Delete...
> * DDL: CREATE TABLE, Drop table ...
> * DAL: SHOW...

Thanks Eric! I'm about to have separate articles for each command. I'll see if we need to group some of them based on the dml/ddl categories if there isn't much to include in a single article.

PR #98, opened 2022-06-01
fix: add missing semicolons

kwannoel:


hengm3467 (approved):
@kwannoel Thanks for fixing it!

PR #110, opened 2022-06-09
Test the formatting of sql commands

hengm3467:
Please do not approve just yet. We are adding more commands.


hengm3467:
@xiangjinwu Please take a look at the SQL syntax style in sql-select.md. Let me know what you think. Here is a preview link: https://pr-110.dmfzqghdggtry.amplifyapp.com/docs/latest/intro/

docs/sql/commands/sql-select.md
```diff
@@ -0,0 +1,103 @@
...
+
+Conditions for the ON clause that must be met before the two from_items can be joined.
+
+### `window_type`
+
+The type of the time window function. Possible values are `HOP` and `TUMBLE`.
+
+### `interval_expression`
+
+The interval expression, in the format of `INTERVAL '<count>' <timeunit>`. For example: `INTERVAL '1' DAYS`.
```
xiangjinwu: `INTERVAL '1' DAY` without `S`.

It would be treated as `interval '1' as days` where `days` is the output column alias and unit is implicitly `second`.
Note that `interval '2 days'` works fine with `days` inside quotes, no matter `days` or `day`.

docs/sql/commands/sql-select.md
```diff
@@ -0,0 +1,103 @@
...
+Use the `SELECT` command to retrieve rows from a table or materialized view. It returns the rows that satisfy the creteria that you specify with the clauses and conditions in your query.
+
+
+```sql
+SELECT [ ALL | DISTINCT] [ * | <expression> [ AS <output_name> ] [, <expression> [ AS <output_name> ]...] ]
+    [ FROM <from_item> [, <from_item> ...] ]
+    [ WHERE <condition> ]
+    [ GROUP BY <grouping_expression> [, <grouping_expression>... ]
+    [ HAVING <condition> ]
+    [ ORDER BY <ordering_expression> [ ASC | DESC ] [ NULLS { FIRST | LAST } ] [, ...] ]
```
xiangjinwu: We do not support specifying `NULLS { FIRST | LAST }` yet. Moreover, our current default behavior is actually the opposite of PostgreSQL.

docs/sql/commands/sql-select.md
```diff
@@ -0,0 +1,103 @@
...
+    [ LIMIT { count | ALL } ]
+    [ OFFSET start [ ROW | ROWS ] ]
+```
+Where `from_item` can be:
+```sql
+    <table>  [ [ AS ] <alias> [ ( <column_alias_list> ) ]]
+    <window_type> <table> <col> <interval_expression> [ [ AS ] <alias> [ ( <column_alias_list> ) ] ]
```
xiangjinwu: `<window_type> (<table>, <col>, <interval_expression>)`
with parentheses and commas

docs/sql/commands/sql-select.md
```diff
@@ -0,0 +1,103 @@
...
+    <window_type> <table> <col> <interval_expression> [ [ AS ] <alias> [ ( <column_alias_list> ) ] ] 
+    (<select>) [ [ AS ] <alias> [ ( <column_alias_list> ) ] ] 
+    <from_item>  <join_type> <from_item> [ ON <join_condition> ]
+```
+
+## Parameters
+
+### `alias`
+
+An alias is a temporary alternative name for columns, tables, views, materialized views, etc. in a query.
```
xiangjinwu: In the sqls above `<alias>` always refers to tables or materialized views but never columns. Column aliases are referred to as `<output_name>` or `<column_alias_list>` above.

docs/sql/commands/sql-select.md
```diff
@@ -0,0 +1,103 @@
...
+
+The name of a table or materialized view.
+
+### `expression`
+
+A column or an expression.
+
+### `from_item`
+
+It can be a table, materialized view, subquery, or join clause.
```
xiangjinwu: Already have more detailed explanation above, which includes table functions.

docs/sql/commands/sql-select.md
```diff
@@ -0,0 +1,103 @@
...
+
+### GROUP BY clause
+
+`grouping_expression` can be:
+- Input column names
+- Input column expressions without subqueries or correlated columns
+- Output column names
```
xiangjinwu: GROUP BY does not support `Output column names` (or output column ordinal numbers) yet.

docs/sql/commands/sql-select.md
```diff
@@ -0,0 +1,103 @@
...
+```
+Where `from_item` can be:
+```sql
+    <table>  [ [ AS ] <alias> [ ( <column_alias_list> ) ]]
+    <window_type> <table> <col> <interval_expression> [ [ AS ] <alias> [ ( <column_alias_list> ) ] ] 
+    (<select>) [ [ AS ] <alias> [ ( <column_alias_list> ) ] ] 
+    <from_item>  <join_type> <from_item> [ ON <join_condition> ]
+```
+
+## Parameters
```
xiangjinwu, line 30: Subtitles in `Parameters` are a little bit inconsistent: some are referring to placeholders in sqls above (e.g. `<alias>`, `<ordering_expression>`) while others are referring to clauses (e.g. `GROUP BY`, `LIMIT`).

Which style do we prefer?

docs/sql/commands/sql-select.md
hengm3467: @xiangjinwu Thanks. To summarize, the default time unit is seconds, and we distinguish plural and singular forms. If the count is 1, then the time unit must be in the singular forms (for example, "DAY", "HOUR", "MINUTE", and "SECOND". Correct?

And we expect people to write as "interval '1' day" or "interval '1 day'"? That is, time unit should be inside or outside quotes? Thanks.

docs/sql/commands/sql-select.md
hengm3467: @xiangjinwu Does it mean that we use DESC as the default ordering option? Can you clarify on the "opposite" part? :)  Thanks.

docs/sql/commands/sql-select.md
xiangjinwu: Using singular vs plural is actually not related to whether it is 1 or 2 here.

There are 2 ways to write an interval literal: with a unit outside quotes, or without units outside quotes.
* It is standard SQL to use a unit outside quotes. And it must be singular.
* It is PostgreSQL extension to allow units inside quotes. It accepts both singular and plural, and default to `second` when no unit is specified (`interval '4'`).

docs/sql/commands/sql-select.md
xiangjinwu: |direction|null|RW|PG|
|-|-|-|-|
|ASC or unspecified|NULLS FIRST|NotSupported|null-small-large|
|ASC or unspecified|NULLS LAST|NotSupported|small-large-null|
|ASC or unspecified|**unspecified**|null-small-large|small-large-null|
|DESC|**unspecified**|large-small-null|null-large-small|
|DESC|NULLS FIRST|NotSupported|null-large-small|
|DESC|NULLS LAST|NotSupported|large-small-null|

* When `asc`, **unspecified** means `nulls last` in PG but `nulls first` in RW.
* When `desc`, **unspecified** means `nulls first` in PG but `nulls last` in RW.

docs/sql/commands/sql-select.md
hengm3467: Some parameters are used in more than one clauses, so it would complicate the layout if we want to list parameters under clauses. We can tweak the layout in the future. :)

CharlieSYH (approved):

PR #118, opened 2022-06-21
Added more CREATE and SHOW commands

emile-00:


docs/sql/commands/sql-create-schema.md
```diff
@@ -0,0 +1,25 @@
...
+description: Create a new schema.
+slug: /create-schema
+---
+
+Use the `CREATE SCHEMA` command to create a new schema.
+
+## Syntax
+
+```sql
+CREATE SCHEMA IF NOT EXISTS [database.]<schema>;
```
CharlieSYH: "IF NOT EXISTS" is optional. Please see the [e2e test record](https://github.com/singularity-data/risingwave/blob/main/e2e_test/ddl/schema.slt) here.
Suggestion: CREATE SCHEMA [ IF NOT EXISTS ] [database.]<schema>;

docs/sql/commands/sql-create-schema.md
```diff
@@ -0,0 +1,25 @@
...
+```
+## Parameters
+|Parameter or clause        | Description           |
+|---------------------------|-----------------------|
+|*schema*                   |The name of the schema to be created.|
+|<b>IF NOT EXISTS</b> *clause*      |Creates a schema if the schema name has not already been used. Otherwise throws an error.|
```
CharlieSYH: Since the word "clause" is not a parameter, we do not have to italicize it.
Suggestion: drop the * *.
And you can use ** IF NOT EXISTS ** to make it bold. It's easier.

docs/sql/commands/sql-create-schema.md
```diff
@@ -0,0 +1,25 @@
...
+|*database*                   |The name of the database for the schema to be created in.|
```
CharlieSYH: I think it is better if we tell users what's the default database (I think it is called "dev" but not sure) if they don't specify a certain database.
Or we don't if such info is redundent. Your call.

docs/sql/commands/sql-create-schema.md
```diff
@@ -0,0 +1,25 @@
+---
+id: create-schema
```
CharlieSYH: New [rules for the front matter](https://www.notion.so/singularity-data/SQL-command-template-e14b5ceeb5474c0baac05c67e49162f1): id and slug should start with "sql-" for SQL commands.
Suggestion: sql-create-schema (id and slug); and check the other files. Thanks!

docs/sql/commands/sql-show-tables.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Syntax
+
+```sql
+SHOW TABLES [FROM <schema>];
+```
+## Parameters
+|Parameter   | Description           |
+|---------------------------|-----------------------|
+|*schema*                   |Schema from which tables will be listed. If not given, tables from the default schema will be listed.|
```
CharlieSYH: Which better describe the reality, from or in a schema? Google tells me the latter is more common.

CharlieSYH (approved):

docs/sql/commands/sql-show-mv.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Syntax
+
+```sql
+SHOW MATERIALIZED VIEWS [FROM <schema>];
+```
+## Parameters
+|Parameter      | Description           |
+|---------------------------|-----------------------|
+|*schema*                   |The schema from which the materialized views will be listed. The default schema is public.|
```
CharlieSYH: "public"?

hengm3467 (approved):

PR #127, opened 2022-06-30
Added examples

emile-00:
Added examples for select, show databases, and show schemas. Would appreciate feedback and suggestions on the setup of the example for the select command and the complexity of the query. Will add more examples of queries based on suggestions!


CharlieSYH:
Love the query sample for SELECT. I think we can create a whole bunch of examples that fit in the taxi company narrative.
BTW, do you think we should add what's returned from RisingWave to the examples of certain commands? That will make commands like SHOW and SELECT more informative. For example: 
```sql
 SHOW DATABASES;
```
```
 Name 
------
 dev
 taxi
(2 rows)
```

CharlieSYH (approved):

hengm3467 (approved):

PR #137, opened 2022-07-05
fix docker-compose error

### LINKED ISSUE #134: docker version
In the ad-ctr tutorial,
>docker-compose up -d
ERROR: The Compose file './docker-compose.yml' is invalid because:
'name' does not match any of the regexes: '^x-'
>You might be seeing this error because you're using the wrong Compose file version. Either specify a supported version (e.g "2.2" or "3.3") and place your service definitions under the `services` key or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/

See the issue at https://github.com/docker/compose/issues/9412

We may need to either remind users of the docker's version or ask users to enable "Use Docker Compose V2" in Docker Desktop to fix the problem.

lmatz:
We may also just keep `docker compose` instead.

closes #134

### DIFF
--- a/docs/tutorials/real-time-ad-performance-analysis.md
+++ b/docs/tutorials/real-time-ad-performance-analysis.md
@@ -42,6 +42,15 @@ cd ad-ctr
 docker-compose up -d
 ```
 
+If error occurs as below:
+```shell
+ERROR: The Compose file './docker-compose.yml' is invalid because:
+'name' does not match any of the regexes: '^x-'
+```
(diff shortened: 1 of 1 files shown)


CharlieSYH (approved):
Reviewed and made some minor changes. Merge if there're no further changes.

lmatz:
> Reviewed and made some minor changes. Merge if there're no further changes.

Thanks! I have no further changes

PR #153, opened 2022-07-13
fix: not only x86-64 can run the image

neverchanje:


skyzh:
IIRC we still only have x86 image for singularity-data/risingwave? The arm image is built as risingwave-ARM64-testing manually, and is not part of the CI pipeline.

PR #154, opened 2022-07-13
Revamp operators & funtions

CharlieSYH:


docs/sql/functions-operators/sql-function-logical.md
```diff
@@ -0,0 +1,25 @@
...
+
+| Operator | Expression & Description |
+| ----------- | ----------- | 
+| AND | `boolean1 AND boolean2` <br /> Logical AND. <br />  TRUE if both *boolean1* and *boolean2* are TRUE. |
+| OR | `boolean1 OR boolean2` <br /> Logical OR. <br /> TRUE if either *boolean1* or *boolean2* is TRUE. |
+| NOT | `NOT boolean` <br /> Negates value. <br /> |
+
+**Example**
+
+| NOT a | a | b | a AND b | a OR b |
```
emile-00: I really like the idea of the table as a concise way to display how all the operators work! But I think it could be clarified if 'a' and 'b' columns are put first since they're the givens and the rest of the operators depend on those.

docs/sql/functions-operators/sql-function-string.md
```diff
@@ -0,0 +1,47 @@
...
+slug: /sql-function-string
+title: String functions and operators
+---
+
+
+## String functions
+
+|Function|Description|Example|
+|---|---|---|
+|REPLACE (string varchar, from varchar, to varchar) → varchar|Replaces all occurrences in string of substring from with substring to.|`REPLACE('abcdefabcdef', 'cd', 'XX')` → `abXXefabXXef`|
```
emile-00: Maybe: Replaces all occurrences in string of substring 'from' with substring 'to'.
It's a little bit difficult to read without highlighting that.

docs/sql/functions-operators/sql-function-string.md
```diff
@@ -0,0 +1,47 @@
...
+|TRIM (string varchar) → varchar|Removes the longest string containing only spaces from the start and end of string.|`TRIM(' cake   ')` → `cake`|
+|LTRIM (string varchar) → varchar|Removes the longest string containing only spaces from the start of string.|`LTRIM('   cake')` → `cake`|
```
emile-00: I think using the same '  cake  ' as the example input string would be helpful to show that LTRIM and RTRIM only remove spaces from one side of the string.

docs/sql/functions-operators/sql-function-string.md
```diff
@@ -0,0 +1,47 @@
...
+|RTRIM (string varchar) → varchar|Removes the longest string containing only spaces from the end of string|`RTRIM('cake   ')` → `cake`|
+|SUBSTR (string varchar, start integer [, count integer ]) → varchar|Extracts the substring of string starting at the start-th character, and extending for count characters if that is specified.| `SUBSTR('alphabet', 3)` → `phabet`; <br /> `SUBSTR('alphabet', 3, 2)` → `ph`|
```
emile-00: Maybe: Extracts the substring from 'string' starting from the letter at the given integer value, and extending for 'count' characters, if specified.
Do we support negative start integers?

docs/sql/functions-operators/sql-function-string.md
```diff
@@ -0,0 +1,47 @@
...
+|UPPER (varchar) → varchar|Converts the string to all upper case.|`UPPER('tom')` → `TOM`|
```
emile-00: Uppercase and lowercase are one word

docs/sql/functions-operators/sql-function-string.md
```diff
@@ -0,0 +1,47 @@
...
+|LOWER (varchar) → varchar|Converts the string to all lower case.|`LOWER('TOM')` → `tom`|
+|POSITION (string varchar, substring varchar) → integer	|Returns first starting index of the specified substring within string, or zero if it is not present.|`POSITION('high', 'ig')` → `2`|
```
emile-00: Can remove 'first' from description

docs/sql/functions-operators/sql-function-string.md
CharlieSYH: I italicized them:
Replaces all occurrences of substring *from_string* in *string* with substring *to_string*.

docs/sql/functions-operators/sql-function-string.md
CharlieSYH: That's just brilliant.

docs/sql/functions-operators/sql-function-string.md
CharlieSYH: Unlike MySQL, RisingWave and PG don't support a negative *start*.
See the query:
```
dev=> select SUBSTR('alphabet', -3);
 ?column? 
----------
 alphabet
(1 row)

dev=> select SUBSTR('alphabet', -1);
 ?column? 
----------
 alphabet
(1 row)
```
My revision:
Extracts the substring from *string* starting at position *start* and extending for *count* characters, if specified. *start* should be equal to or larger than 1.

docs/sql/functions-operators/sql-function-string.md
CharlieSYH: : Returns the starting index of the specified *substring* within *string*...

docs/sql/functions-operators/sql-function-logical.md
CharlieSYH: :
| a | b | a AND b | a OR b | NOT a | 
| - | - | ------- | ------ | ----- |
| TRUE | TRUE | TRUE | TRUE | FALSE |
| TRUE | FALSE | FALSE | TRUE | FALSE |
| TRUE | NULL | NULL | TRUE | FALSE |
| FALSE | FALSE | FALSE | FALSE | TRUE |
| FALSE | NULL | FALSE | NULL | TRUE |
| NULL | NULL | NULL | NULL | NULL |

emile-00 (approved):

hengm3467 (approved):

PR #156, opened 2022-07-13
doc updates about struct

hengm3467:
Add "struct" to the list of data types, and also to the "CREATE TABLE" command page. The syntax does not reflect the nested structure yet, as we use angle brackets to represent parameters and nested tables are enclosed with angle brackets. It would be confusing if we include nested tables in the syntax at the moment. We'll customize the styling of the syntax blocks, and will add nested tables to the syntax when it's ready.


CharlieSYH:
Working on it. We need to use a new styling rule for the code block  so that we could ditch the current practice of wrapping parameter name with <>.
See and edit the draft here: https://www.icloud.com.cn/keynote/057rFyhP_Ljr3jstqHxv8lOMA#Untitled

CharlieSYH (approved):

PR #157, opened 2022-07-14
example of recursive struct

lmatz:
Just a nit.
My first thought when I see the example is wondering whether `struct` can be used recursively within another `struct`.
Try to maximize the info conveyed in one example.


hengm3467 (approved):
Good addition!

PR #168, opened 2022-07-19
update: syntax change for with clause

lmatz:
Due to the syntax change introduced by https://github.com/singularity-data/risingwave/pull/3997


CharlieSYH:
Need to change the syntax for CREATE SOURCE related articles like https://www.risingwave.dev/docs/latest/create-source-kafka-redpanda/

lmatz:
> Need to change the syntax for CREATE SOURCE related articles like https://www.risingwave.dev/docs/latest/create-source-kafka-redpanda/

Thanks for pointing out. All of 4 are updated.

hengm3467 (approved):
Looks good to me. Thanks!

PR #171, opened 2022-07-21
updated x64 and ARM64 instructions

hengm3467:
I updated the chipset-related instructions based on the discussion. Please let me know if these are accurate.

### DIFF
--- a/docs/Get-Started.md
+++ b/docs/Get-Started.md
@@ -49,7 +49,7 @@ You can run RisingWave in these ways:
 
 ### Install and run from a Docker image (Linux & macOS)
 
-You can install and run RisingWave from a Docker image.
+You can install and run RisingWave from a Docker image on x64 systems. Images for ARM64 systems (including macOS devices with Apple M1 chips) might be available for testing purpose, but it is not guaranteed.
 
 Ensure you have Docker Desktop installed on your machine. For installation instructions, see [Get Docker](https://docs.docker.com/get-docker/).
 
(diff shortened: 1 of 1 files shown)


docs/Get-Started.md
```diff
@@ -49,7 +49,7 @@ You can run RisingWave in these ways:
 
 ### Install and run from a Docker image (Linux & macOS)
 
-You can install and run RisingWave from a Docker image. Currently, only x86-64 platforms are supported.
+You can install and run RisingWave from a Docker image on x64 systems. Images for ARM64 systems (including macOS devices with Apple M1 chips) might be available for testing purpose, but it is not guaranteed.
```
skyzh: x64 -> x86-64 or amd64

skyzh (approved):
Rest LGTM

docs/Get-Started.md
hengm3467: @skyzh As I understand it, x64 covers both x86_64 and amd64. Do we want to single either of them out?

skyzh: I think either of x86-64 and amd64 would be okay here. I seldom see x64...

hengm3467: or we can keep referring to it as x86-64.

hengm3467: > I think either of x86-64 and amd64 would be okay here. I seldom see x64...

Cool. I changed it to AMD64.

PR #229, opened 2022-08-17
Adding the cast function and the mapping table

### LINKED ISSUE #128: Update "data types" article to include information about casts
Summary: Support casts from number types (smallint, integer, bigint, numeric, real, and double precision) to varchar using the to_string method. 
Related RisingWave PR is [#2522. ](https://github.com/singularity-data/risingwave/pull/2522)

CharlieSYH:
## Info
- **Description**: 
Adding the cast function & operator and the mapping table
See the preview here: https://pr-229.dmfzqghdggtry.amplifyapp.com/docs/latest/sql-data-types/#casts

- **Related code PR**: 
https://github.com/singularity-data/risingwave/pull/2522

- **Related doc issue**: 
Resolves https://github.com/singularity-data/risingwave-docs/issues/128

- **Notes**: 
Cast mapping info: https://github.com/singularity-data/risingwave/blob/be868cc6e479de30be78c98b77ab3ad686938b89/src/frontend/src/expr/type_inference/cast.rs#L201

## Before merging
  - [x] (For version-specific PR) I have selected the corresponding software version in **Milestone** and linked the related doc issue to this PR in **Development**.
  - [x] I have checked the doc site preview and the updated parts look all good. <details><summary>How?</summary>image</details>
  - [x] (For version-specific PR) I have confirmed that this PR is for a released version. If the software version is not released, put it on hold.

## Published doc pages
  [ After merging this PR, edit the description to include the links to the updated doc pages here. For example, https://www.risingwave.dev/docs/latest/intro/. ]


xiangjinwu:
Some suggestions on the appearance:
* The cast table is wider than my screen and seems it cannot scroll
* Maybe we can also use icons for implicit/assign/explicit in the table, just like ❌

CharlieSYH:
> * The cast table is wider than my screen and seems it cannot scroll

Opened an issue on it: https://github.com/singularity-data/risingwave-docs/issues/230

CharlieSYH:
> * The cast table is wider than my screen and seems it cannot scroll
> * Maybe we can also use icons for implicit/assign/explicit in the table, just like ❌

Both comments have been resolved. @xiangjinwu @hengm3467 Please review again. https://pr-229.dmfzqghdggtry.amplifyapp.com/docs/latest/sql-data-types/#casts

hengm3467 (approved):
Looks good to me. We can consider it the first iteration. We might be able to improve the icons used later on.

xiangjinwu (approved):
Thanks for the improvements! It looks good enough now.

Just one more thing if it can be achieved easily: Is it possible to freeze the 1st column (just like the 1st row) when scrolling?

CharlieSYH:
> Just one more thing if it can be achieved easily: Is it possible to freeze the 1st column (just like the 1st row) when scrolling?

We actually planned to do that, but it conflicts with the sticky header feature (freezing the 1st row). It would take extra time from the web team, so we can implement that in the future when we have more tables like this one. @xiangjinwu

CharlieSYH:
> We might be able to improve the icons used later on.
We can replace them if you find more suitable ones. Emojis might not be very suitable for tech docs. @hengm3467

PR #368, opened 2022-11-17
feat: add LinkedIn Icon

TingJianZheng:
feat: add LinkedIn Icon


cloudcarver (approved):
LGTM!

CharlieSYH:
Thanks @TingJianZheng for taking care of this.

Seems that we still have some ratio issues to be solved before merging this PR:
Some icons are more narrow than they should be. I think it has something to do with the ratio settings (height/width).
Portrait (mobile):
image
Landscape (wide-screen):
image

And the icons are not perfectly aligned with the back-to-top button:
image

TingJianZheng:
> Thanks @TingJianZheng for taking care of this.
> 
> Seems that we still have some ratio issues to be solved before merging this PR: Some icons are more narrow than they should be. I think it has something to do with the ratio settings (height/width). Portrait (mobile): image Landscape (wide-screen): image
> 
> And the icons are not perfectly aligned with the back-to-top button: image

I think it's the padding, I changed the way the icon is centered

CharlieSYH:
Big thanks! Perfect now.

CharlieSYH (approved):

PR #421, opened 2022-12-06
document the rank() function in Top-N queries

### LINKED ISSUE #385: Document TopN with ties for batch
### Related code PR

https://github.com/risingwavelabs/risingwave/pull/6383

The Top-N topic needs to be added back to the sidebar under a new directory "SQL patterns". 

### Reference

https://sqlsunday.com/2013/09/01/using-top-n-with-ties/

hengm3467:
## Info
- **Description**: 
Updated on Dec 8: 
Revert the topic to be about topN only and added the rank() function.

- **Related code PR**: 
Releated tracking issues is: https://github.com/risingwavelabs/risingwave/issues/4847

- **Related doc issue**: 
Resolves #385 

- **Notes**: 

## Before merging
</details>

## Published doc pages
  [ After merging this PR, edit the description to include the links to the updated doc pages here. For example, https://www.risingwave.dev/docs/latest/intro/. ]


docs/sql/sql-window-functions.md
```diff
@@ -0,0 +1,52 @@
+---
+id: sql-window-functions
+slug: /sql-window-functions
+title: Window functions
+---
+
+A window function performs a calculation across a set of table rows that are related to the current row, and returns a single result for each row. It is different from an aggregate function, which returns a single result for a group of rows.
+
+A window function includes an OVER clause, which defines a window of rows around the row being evaluated.
```
CharlieSYH: I think here we should use the code format: `OVER` clause

xxchan (commented):
Actually we haven't implement the general-purpose OVER window functions yet. BTW there's an RFC https://github.com/risingwavelabs/rfcs/pull/8.

> One thing to confirm is where these functions can be used. Only in FROM clauses?

The `ROW_NUMBER()` and `RANK()` can only be written as the special pattern described in the top-n queries doc.

Just like flink says: [Note: the above pattern must be followed exactly, otherwise the optimizer won’t be able to translate the query.
](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/sql/queries/topn/#:~:text=Note%3A%20the%20above%20pattern%20must%20be%20followed%20exactly%2C%20otherwise%20the%20optimizer%20won%E2%80%99t%20be%20able%20to%20translate%20the%20query)

Therefore, I think we should not document window functions now. Even after we support it, we'd better keep the doc for TopN queries due to the query optimization and their frequent usage.

> Also, I don't see an example for `WITH TIES`

Actually I'm considering to ban it. I just implemented it because it's closed related to `RANK()`, and not hard to implement. But it turns out to have significant difference with `RANK()`.

docs/sql/syntax/sql-queries-top-n.md
```diff
@@ -1,31 +0,0 @@
...
-
-:::note
-
-`rank` cannot be included in `column_list`.
-
-:::
-
-|Parameter|Description|
-|---|---|
-|`ROW_NUMBER()`|Assigns a unique, sequential number to each row, starting with one, according to the ordering of rows within the partition. `RANK()` is supported for materialized views but not tables.|
```
xxchan, line 28: > `RANK()` is supported for materialized views but not tables

This is indeed outdated (and actually was wrong..)

docs/sql/sql-window-functions.md
```diff
@@ -0,0 +1,52 @@
...
+
+RisingWave now supports two general-purpose window functions:
+
+- `row_number()`: Returns the number of the current row within its partition, counting from 1.
+- `rank()`: Returns the ordinal (1-based) rank of the current row within its partition, with gaps. All peer rows receive the same rank value. The next row or set of peer rows receives a rank value which increments by the number of peers with the previous rank value.
```
xxchan: Can be used to improve the TopN doc

docs/sql/syntax/sql-queries-top-n.md
hengm3467: Noted. I have removed this sentence.

hengm3467:
@xxchan I reverted the topic to be only about topN. Please review and add your comments if any. Thanks!

docs/sql/syntax/sql-pattern-topn.md
```diff
@@ -0,0 +1,50 @@
+---
+id: sql-pattern-topn
+slug: /sql-pattern-topn
+title: Top N by group
+---
+
+Top-N queries are queries that limit the result to a specific number of rows. Top N queries are helpful to display only the N top-most or the N bottom-most records from a table or view based on a condition.
```
xxchan: Can be more precise like this?

```suggestion
Top-N queries return only the N top-most or the N bottom-most records from a table or view based on a condition.
```

Also, maybe we should use either `Top N` or `Top-N` to be more consistent.

docs/sql/syntax/sql-pattern-topn.md
```diff
@@ -0,0 +1,50 @@
...
+
+:::info
+
+You must follow the pattern exactly to construct a valid Top-N query.
+
+:::
+
+|Parameter|Description|
+|---|---|
+|*function_name*| RisingWave supports two functions in Top-N queries: <br />`row_number()`: Returns the number of the current row within its partition, counting from 1.<br />`rank()`: Returns the ordinal (1-based) rank of the current row within its partition, with gaps. All peer rows receive the same rank value. The next row or set of peer rows receives a rank value which increments by the number of peers with the previous rank value.|
```
xxchan: > Returns the number of the current row within its partition

"the number" seems not very clear. 

Also, is it possible to rephrase the descriptions for `rank` and `row_number` using similar sentences. That would be easier for users to understand 🤔

docs/sql/syntax/sql-pattern-topn.md
```diff
@@ -0,0 +1,50 @@
...
+|`PARTITION BY` clause |Specifies the partition columns. Each partition will have a Top-N result.|
+|`ORDER BY` clause|Specifies the ordering columns.|
+|`WHERE rank <= N`|Required for the query to be recognized as a Top-N query. You can specify an additional condition to further narrow down the scope of results. For example: `WHERE rank <=N AND rank > M`|
```
xxchan:
```suggestion
|`WHERE rank <= N`|Required for the query to be recognized as a Top-N query.|
|`[AND conditions]`|You can specify any additional conditions to further filter the results. |
```

For `rank <=N AND rank > M`, it's actually not supported for `rank()` yet, but only `row_number()`. Also I guess if the user has the need, they may just try it to see whether it works.. The syntax is natural. So I'm not sure whether we want to document it. 

If we want to document it, I think it should be sth like:
```
|`WHERE rank <= N`| Specifies the range of the rank number. Required for the query to be recognized as a Top-N query. It can also be written as other forms of ranges like: `WHERE M < rank AND rank < N` or `WHERE rank between M and N` |
```

xxchan (commented):
Others LGTM

docs/sql/syntax/sql-pattern-topn.md
```diff
@@ -0,0 +1,51 @@
...
+
+
+## Syntax
+
+```sql
+SELECT [column_list] 
+  FROM (
+    SELECT [column_list], 
+      function_name() OVER ([PARTITION BY col1[, col2...]] 
+        ORDER BY col1 [ ASC | DESC ][, col2 [ ASC |DESC ]...]) AS rank
```
CharlieSYH: Missing a space before DESC `col2 [ ASC |DESC ]`

docs/sql/syntax/sql-pattern-topn.md
```diff
@@ -0,0 +1,51 @@
...
+
+:::info
+
+You must follow the pattern exactly to construct a valid Top-N query.
+
+:::
+
+|Parameter|Description|
+|---|---|
+|*function_name*| RisingWave supports two functions in Top-N queries: <br />`row_number()`: Returns the sequential row ordinal (1-based) of each row for each ordered partition. <br />`rank()`: Returns the ordinal (1-based) rank of each row within the ordered partition. All peer rows receive the same rank value. The next row or set of peer rows receives a rank value which increments by the number of peers with the previous rank value.|
```
CharlieSYH: Consider using a bullet list for the two functions: <ul>...</ul>

CharlieSYH (approved):

PR #567, opened 2023-02-16
Document internal tables

hengm3467:
## Info
- **Description**: 
Create the reference information about `SHOW INTERNAL TABLES`. Added a bit introduction to internal tables and internal states. 

Later on, we may need to have a how-to guide about how to use internal tables for performance tuning or other possible scenarios. This will be done after we have the new articles ready by Tenny, and will be included in a separate PR.

- **Preview**: 
[ Paste the preview link to the edited page here. ]

- **Related code PR**: 
[ Provide the link to the code PR here. For example, https://github.com/risingwavelabs/risingwave/pull/4085. Delete this part if there's no code PR related. ]

- **Related doc issue**: 
Resolves [ Provide the link to the doc issue here. ]

- **Notes**: 
[ Any additional information? ]

## Before merging
  - [ ] (For version-specific PR) I have selected the corresponding software version in **Milestone** and linked the related doc issue to this PR in **Development**.
  - [ ] I have acquired the approval from the owner (and optionally the reviewers) of the code PR and at least one tech writer (`bernscode`, `CharlieSYH`, `emile-00`, & `hengm3467`). 
  - [ ] I have checked the doc site preview, and the updated parts look good. <details><summary>How?</summary>Scroll down and open this link: image
</details>

PR #569, opened 2023-02-16
Revert "Document internal tables"

hengm3467:
Reverts risingwavelabs/risingwave-docs#567


TennyZhuang:
Either reverting or adding a warning message is acceptable to me, any ideas @neverchanje ?

hengm3467:
@TennyZhuang @st1page For your information, I discussed with @neverchanje , and we agreed not to revert this PR, but instead add a note and move it to a directory that makes more sense (under 'Manage RisingWave')

PR #590, opened 2023-02-22
added redash as an available "BI and data analytic platforms"

cameronbraid:
since redash works, its worth showing this in the docs

### DIFF
--- a/versioned_docs/version-0.1.16/rw-integration-summary.md
+++ b/versioned_docs/version-0.1.16/rw-integration-summary.md
@@ -98,7 +98,7 @@ For tools or integrations that you would like to use but are not listed in the t
 |Jupyter Notebook| |Researching <voteNotify note="jupyter" />|
 |Looker| |Researching <voteNotify note="looker" /> |
 |Metabase | |In progress <voteNotify note="metabase" />|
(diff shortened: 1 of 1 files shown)


versioned_docs/version-0.1.16/rw-integration-summary.md
```diff
@@ -98,7 +98,7 @@ For tools or integrations that you would like to use but are not listed in the t
 |Jupyter Notebook| |Researching <voteNotify note="jupyter" />|
 |Looker| |Researching <voteNotify note="looker" /> |
 |Metabase | |In progress <voteNotify note="metabase" />|
-
+|https://redash.io/ || Available |
```
hengm3467: Thanks! @cameronbraid I would like to suggest a minor change. Maybe just "Redash", to be consistent with other systems. 

The initial reason we didn't use hyperlinked names is it would make the whole page too busy with links. This may change, and if that happens, we can do a global change by adding links.

cameronbraid: sure thing

hengm3467 (approved):
@cameronbraid Thanks a lot!

PR #600, opened 2023-02-24
Document the parsing of unknown typed literal

### LINKED ISSUE #546: Document the handling of unsupported cast expressions
### Related code PR

https://github.com/risingwavelabs/risingwave/pull/7378

Unsupported cast expressions will be treated as literal expressions. We need to add a note in the topic about casting operators.

### Reference

_No response_

hengm3467:
## Info
- **Description**: 
Casting expressions that cannot be parsed will be treated as strings literally. Add a note about this approach. 

- **Preview**: 
[ Paste the preview link to the edited page here. ]

- **Related code PR**: 
https://github.com/risingwavelabs/risingwave/pull/7378

- **Related doc issue**: 
Resolves #546 

- **Notes**: 

## Before merging
</details>

### DIFF
--- a/docs/sql/data-types/data-type-casting.md
+++ b/docs/sql/data-types/data-type-casting.md
@@ -25,6 +25,14 @@ Certain data types can be cast to and from other types implicitly or explicitly.
 | **time** | ✖️ | ✖️ | ✖️ | ✖️ | ✖️ | ✖️ | ✖️ | 🟠 | ✖️ | ✖️ | ✖️ |  | 🟢 |
 | **interval** | ✖️ | ✖️ | ✖️ | ✖️ | ✖️ | ✖️ | ✖️ | 🟠 | ✖️ | ✖️ | ✖️ | 🟠 |  |
 
+
+:::note
+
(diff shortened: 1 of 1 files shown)


xiangjinwu:
Sorry for the confusion. Let me elaborate on the "user-facing" part of that PR.

Usually PRs titled "refactor" are not visible to users, and do not need a doc update.

That PR was tagged as "user-facing" maybe due to its intension to provide better user experience in a special case:
```
'2023-02-01 12:00:00'::timestamp with time zone
'2023-02-01 12:00:00+01:00'::timestamp with time zone
```

The first expression is "problematic": it refers to a different real world instant in different contexts, and we want to warn our user when they do this. The second expression is unambiguous and the preferred form.

Before that PR, we were unable to differentiate the 2 cases and would always warn the user conservatively (which is annoying). After that PR, we are smarter and only warn for the first case. This is the only "visible" change to users of the PR.

In conclusion, I do not feel this needs a dedicated doc note.

hengm3467:
@xiangjinwu Thanks for the clarification! I'm closing this PR now.

PR #610, opened 2023-03-01
add new topic about viewing and adjusting system parameters

### LINKED ISSUE #578: Document SHOW PARAMETERS and ALTER SYSTEM
### Related code PR

https://github.com/risingwavelabs/risingwave/pull/7882
https://github.com/risingwavelabs/risingwave/pull/7913

Add two commands SHOW PARAMTERS and ALTER SYSTEM. In SHOW PARAMTERS, list out the parameters, and provide brief descriptions. 

Some very initial description: https://www.notion.so/risingwave-labs/System-Parameters-b9b9c00308bc4020a4208fa1764b24db?pvs=4

### Reference

https://www.postgresql.org/docs/current/config-setting.html

hengm3467:
## Info
- **Description**: 
As described in title. 

- **Preview**: 
Preview link: https://pr-610.d2fbku9n2b6wde.amplifyapp.com/docs/upcoming/view-adjust-system-parameters/

- **Related code PR**: 
risingwavelabs/risingwave#7882
risingwavelabs/risingwave#7913

- **Related doc issue**: 
Resolves  #578 

- **Notes**: 

## Before merging
</details>


docs/manage/view-adjust-system-settings.md
```diff
@@ -0,0 +1,55 @@
...
+slug: /view-adjust-system-settings
+---
+
+## What are system settings?
+
+System settings in RisingWave refer to the parameters that advanced users can use to adjust how internal components work in RisingWave.
+
+Currently, these system parameters are availble in RisingWave. Note that only `barrier_interval_ms` and `checkpoint_interval` can be set.
+
+| Parameter           |    Description    |
```
fuyufjh: Recommend to add mutablity (yes/no) and valid value range

docs/manage/view-adjust-system-settings.md
```diff
@@ -0,0 +1,55 @@
...
+|block_size_kb            | The size of each block in bytes in SSTable.|
+|bloom_false_positive     | False positive rate of bloom filter in SSTable.|
+|state_store              | The state store URL. |
+|data_directory           | The remote directory for storing data and metadata objects.|
+|backup_storage_url       | The URL of the remote storage for backups.|
+|backup_storage_directory | The directory of the remote storage for backups.|
+
+## How to view system settings?
+
+You can use the `SHOW PARAMTERS` statement to view the system parameters and their settings.
```
fuyufjh: Emmm, I thought "system parameters" and "system settings" are synonymous, aren't they?

If so, prefer to use "system parameters" to keep the terminology consistent.

docs/manage/view-adjust-system-settings.md
hengm3467: @fuyufjh Thanks for the comment. "Mutable" or "mutability" is a term that's too technical, and I'm trying to avoid using it. Since there are only two parameters that can be configured by users, I think it's enough to add a note about these two mutable parameters. 
About the value range, I checked the code as suggested by @Gun9niR , it looks like the value ranges for these two parameters are both (1, +infinite). For other parameters, if people cannot set them, why do they need to care about value ranges? :)

docs/manage/view-adjust-system-settings.md
hengm3467: Good suggestion. Let me change to "system parameters".

docs/manage/view-adjust-system-parameters.md
```diff
@@ -0,0 +1,55 @@
...
+## What are system parameters?
+
+System parameters in RisingWave refer to the parameters that advanced users can use to adjust how internal components work in RisingWave.
+
+Currently, these system parameters are availble in RisingWave. Note that only `barrier_interval_ms` and `checkpoint_interval` can be set.
+
+| Parameter           |    Description    |
+|---|---|
+|barrier_interval_ms      | The time interval of the periodic barriers. Its value must be greater than 1.|
+|checkpoint_interval      | Specify the number of barriers for which a checkpoint will be created. Its value must be greater than 1.|
```
Gun9niR: - `barrier_interval_ms` is now immutable 😢 
- Though in the code I wrote `1..`, the actual value range should be any positive integer (maximum value limited by its corresponding type, so it should be `greater than 0`

hengm3467: @Gun9niR Thanks for the clarification. Since there's only one parameter can be configured, do you think we should postpone documenting these two commands? 
Adding @fuyufjh to comment on the way to go. :)

docs/manage/view-adjust-system-parameters.md
```diff
@@ -0,0 +1,55 @@
...
+ sstable_size_mb          | 256
+ block_size_kb            | 64
+ bloom_false_positive     | 0.001
+ state_store              | 
+ data_directory           | hummock_001
+ backup_storage_url       | memory
+ backup_storage_directory | backup
+```
+
+## How to adjust system parameters?
```
Gun9niR, line 47: Maybe we should add a note, that there's no guarantee on the timeliness of param update, similar to flink. We only guarantee that the updated params will be synced to the entire cluster **sooner or later**.

docs/manage/view-adjust-system-parameters.md
Gun9niR: > we should postpone documenting these two commands?

I'm not sure. If the users tend to think "having `ALTER SYSTEM` available but only one param is actually mutable" is lame, we should indeed postpone it 🤣 

But I see no problem to expose `SHOW PARAMETERS`

docs/manage/view-adjust-system-settings.md
fuyufjh: > About the value range, I checked the code as suggested by @Gun9niR , it looks like the value ranges for these two parameters are both (1, +infinite). For other parameters, if people cannot set them, why do they need to care about value ranges? :)

Agree

docs/manage/view-adjust-system-parameters.md
```diff
@@ -0,0 +1,55 @@
...
+
+System parameters in RisingWave refer to the parameters that advanced users can use to adjust how internal components work in RisingWave.
+
+Currently, these system parameters are availble in RisingWave.
+
+| Parameter           |    Description    |
+|---|---|
+|barrier_interval_ms      | The time interval of the periodic barriers.|
+|checkpoint_interval      | Specify the number of barriers for which a checkpoint will be created. The value must be a positive integer.|
+|sstable_size_mb          | There will be a checkpoint for every n barrier.|
```
BugenZhao: The description seems incorrect here?

docs/manage/view-adjust-system-parameters.md
BugenZhao: > https://nightlies.apache.org/flink/flink-docs-release-1.16/docs/dev/table/config/#overview

> Attention: Because options are read at different point in time when performing operations, it is recommended to set configuration options early after instantiating a table environment.

docs/manage/view-adjust-system-parameters.md
hengm3467: Good catch! Fixed.

docs/manage/view-adjust-system-parameters.md
hengm3467: @BugenZhao Thanks for providing the reference link. Quick question: what is the equivalent action in RisingWave to "instantiating a table environment" in Flink? Starting a cluster I guess? Thanks.

BugenZhao: I guess it's beginning to create streaming jobs after launching the cluster. And if we introduce more finer-grained parameters in the future, they may only affect new streaming jobs. The point is we have no guarantee about this. 😂 cc @Gun9niR

hengm3467:
As we already reviewed the content, I think it makes sense that we publish the topic with a note about no guarantee for the effective time. Please take a look and let me know if this is ready for publishing. Thanks.
@Gun9niR

Gun9niR (approved):
LGTM. We should regularly check https://github.com/risingwavelabs/risingwave/blob/main/src/common/src/system_param/mod.rs to keep this doc up to date

hengm3467:
Sounds good. I just added this to the doc release checklist:
https://www.notion.so/risingwave-labs/Release-checklist-for-documentation-7b521df58ab7471b9904900d3a1c4ed6?pvs=4

So we'll check the source code at least once per month to make sure it's accurate. 




> LGTM. We should regularly check https://github.com/risingwavelabs/risingwave/blob/main/src/common/src/system_param/mod.rs to keep this doc up to date

Gun9niR:
Shall we also document the default values of the parameters? I can open a PR to do that.

PR #646, opened 2023-03-14
Document Kafka source upsert

### LINKED ISSUE #634: Document the new upsert parameter and new row formats for Kafka
### Related code PR

https://github.com/risingwavelabs/risingwave/pull/8111

Ingest data -> Kafka, and maybe “CREATE SOURCE”。

### Reference

https://nightlies.apache.org/flink/flink-docs-release-1.16/docs/connectors/table/upsert-kafka/
Above is a page about upsert for Kafka in Flink, as background info about the feature itself. Please evaluate whether and where to include some description of the feature (like its benefits and use scenarios).

emile-00:
## Info
- **Description**: 
Document upsert field for Kafka source
Document upsert json and upsert avro data formats under CREATE SOURCE

- **Preview**: 
[ Paste the preview link to the edited page here. ]

- **Related code PR**: 
https://github.com/risingwavelabs/risingwave/pull/8111

- **Related doc issue**: 
Resolves https://github.com/risingwavelabs/risingwave-docs/issues/634

- **Notes**: 

## Before merging
</details>


xx01cyx (commented):
The PR https://github.com/risingwavelabs/risingwave/pull/8168 is actually about Kafka **sink** instead of source. Besides, the upsert Kafka sink is an unstable feature now and we may wish to postpone documenting this until stable.

cc. @tabVersion

emile-00:
Oops! Sorry about that, I linked the incorrect doc issue. This was meant for the Kafka source upsert updates. @xx01cyx

docs/sql/commands/sql-create-source.md
```diff
@@ -180,9 +180,27 @@ ROW FORMAT DEBEZIUM_JSON
...
+When creating a source from streams in Maxwell JSON, you can define the schema of the source within the parentheses after the source name (`schema_definition` in the syntax), and specify the format in the `ROW FORMAT` section. You can directly reference data fields in the JSON payload by their names as column names in the schema.
 
 Syntax:
 ```sql
 ROW FORMAT MAXWELL
 ```
+
+### Upsert JSON
+
+When consuming data in JSON from Kafka topics, the data format needs to be specified as `UPSERT_JSON`. RisingWave will be aware that the source message contains key fields as primary columns, as well as the Kafka message value field. If the value field of the message is not null, the row will be updated if the key already exists, or updated if the key does not exist. If the value field is null, the row will be deleted.
```
adevday: > "If the value field of the message is not null, the row will be updated if the key already exists, or updated if the key does not exist." 

1. Maybe there is a little confuse about the 'key exists', it could be 'the message key is not empty' or 'the row that referenced by the key exists in the database table'.
2. And it should be 'or **inserted** if the (row referenced by ) key does not exist (in database table)' for my understanding

docs/create-source/create-source-kafka.md
```diff
@@ -155,7 +156,21 @@ WITH (
...
+
+```sql
+CREATE SOURCE IF NOT EXISTS source_abc 
+WITH (
+   connector='kafka',
+   upsert='true',
+   properties.bootstrap.server='localhost:9092',
+   topic='test_topic'
+)
+ROW FORMAT UPSERT_AVRO
```
tabVersion, line 173: @idx0-dev please confirm that we allow schemaless source for `upsert_avro`?

docs/create-source/create-source-kafka.md
```diff
@@ -126,13 +126,14 @@ For materialized sources with primary key constraints, if a new data record with
 |properties.bootstrap.server| Required. Address of the Kafka broker. Format: `'ip:port,ip:port'`.	|
 |scan.startup.mode|Optional. The offset mode that RisingWave will use to consume data. The two supported modes are `earliest` (earliest offset) and `latest` (latest offset). If not specified, the default value `earliest` will be used.|
 |scan.startup.timestamp_millis|Optional. RisingWave will start to consume data from the specified UNIX timestamp (milliseconds). If this field is specified, the value for `scan.startup.mode` will be ignored.|
+|upsert| Optional. If true, RisingWave will read messages from Kafka topics in the upsert fashion.|
```
tabVersion, line 129: please note that upsert mode can only be used together with kafka connector (other mqs don't have key/value fields in a single message)

tabVersion (approved):

docs/create-source/create-source-kafka.md
adevday: Yes, but you need to explicitly specify the primary key column(s), otherwise the upsert will always act as insert.

tabVersion: so we still need a schema here?

adevday: Partial of the schema that contains all primary key column(s) is enough.

docs/create-source/create-source-kafka.md
emile-00: This parameter was added only to the Kafka Ingestion guide, so this shouldn't cause any misconceptions for the readers!

adevday (approved):

hengm3467 (approved):
LGTM. Thanks.

PR #990, opened 2023-06-21
Document supported join types

### LINKED ISSUE #717: Change required: add documents for all join types.
### Related code PR

We currently support NATURAL JOIN in both streaming and batch mode. But for CROSS JOIN, we only support it in the batch mode. It's disallowed in the streaming mode because of the expensive cost of nested-loop join.

Typically, databases have a seperated doc page named 'Joins' that describes the supported join types. 
Examples:
- https://materialize.com/docs/sql/join/
- https://www.cockroachlabs.com/docs/stable/joins.html
- https://crate.io/docs/crate/reference/en/5.2/concepts/joins.html
- https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/sql/queries/joins/

The content mentioned above should be included in this page.

### Reference

_No response_

hengm3467:
## Info

- **Description**

  [ What's changed? Which parts of the docs are affected? ]

- **Notes**

- **Related code PR**

- **Related doc issue**
  
  Resolves #717

## For reviewers

- **Preview**

- **Key points**

## Before merging


docs/sql/query-syntax/query-syntax-join-clause.md
```diff
@@ -0,0 +1,173 @@
...
+
+```sql
+<table_expression> FULL [ OUTER ] JOIN <table_expression> ON <join_conditions>;
+<table_expression> FULL [ OUTER ] JOIN <table_expression> USING (<col_name>, <col_name>, ...);
+<table_expression> NATURAL FULL [ OUTER ] JOIN <table_expression>;
+```
+
+## Windows joins
+
+You can join two sources that have the same watermarks. This is called a window join. In a regular join (that is, a join without time attributes), the join state may grow without restriction. With window joins, the join state size can be kept at a resonable size.
```
lmatz: By `same watermarks`, does it mean that ` - INTERVAL '1' SECOND` part must be the same?

lmatz (approved):
rest LGTM

docs/sql/query-syntax/query-syntax-join-clause.md
```diff
@@ -0,0 +1,173 @@
...
+
+```sql
+CREATE SOURCE t1 (ts1 timestamp with time zone, a1 int, b1 int, WATERMARK for ts1 AS ts1 - INTERVAL '1' SECOND) WITH (
+    connector = 'datagen',
+    datagen.rows.per.second = '10',
+);
+CREATE SOURCE t2 (ts2 timestamp with time zone, a2 int, b2 int, WATERMARK for ts2 as ts2 - INTERVAL '1' SECOND) WITH (
+    connector = 'datagen',
+    datagen.rows.per.second = '10',
+);
```
chenzl25:
```suggestion
CREATE SOURCE s1 (
	id int, 
	value int, 
	ts TIMESTAMP, 
	WATERMARK FOR ts AS ts - INTERVAL '20' SECOND
) WITH (connector = 'datagen');


CREATE SOURCE s2 (
	id int, 
	value int, 
	ts TIMESTAMP, 
	WATERMARK FOR ts AS ts - INTERVAL '20' SECOND
) WITH (connector = 'datagen');
```

docs/sql/query-syntax/query-syntax-join-clause.md
```diff
@@ -0,0 +1,173 @@
...
+```
+
+You can join them with the following statement:
+
+```sql
+SELECT * FROM t1
+JOIN t2
+ON ts1 = ts2 
+AND a1 = a2;
```
chenzl25:
```suggestion
CREATE MATERIALIZED VIEW window_join AS
SELECT s1.id AS id1,
       s1.value AS value1,
       s2.id AS id2,
       s2.value AS value2
FROM TUMBLE(s1, ts, interval '1' MINUTE)
JOIN TUMBLE(s2, ts, interval '1' MINUTE)
ON s1.id = s2.id and s1.window_start = s2.window_start;
```

docs/sql/query-syntax/query-syntax-join-clause.md
```diff
@@ -0,0 +1,173 @@
...
+
+In an interval join, the `interval_condtion` must be a watermark-based range.
+
+For example, for sources `t1` and `t2` used in the above section, you can create an interval join:
+
+```sql
+SELECT * FROM t1
+JOIN t2 
+WHERE a1 = a2 
+AND ts1 BETWEEN ts2 - INTERVAL '10' SECOND AND ts2 + INTERVAL '15' SECOND;
```
chenzl25:
```suggestion
CREATE MATERIALIZED VIEW interval_join AS
SELECT s1.id AS id1,
       s1.value AS value1,
       s2.id AS id2,
       s2.value AS value2
FROM s1 JOIN s2
ON s1.id = s2.id and s1.ts between s2.ts and s2.ts + INTERVAL '1' MINUTE;
```

docs/sql/query-syntax/query-syntax-join-clause.md
```diff
@@ -0,0 +1,173 @@
...
+
+The syntax of an interval join is:
+
+```sql
+<table_expression> JOIN <table_expression> ON <equality_join_condition> AND <interval_condition> ...;
+```
+
+In an interval join, the `interval_condtion` must be a watermark-based range.
+
+For example, for sources `t1` and `t2` used in the above section, you can create an interval join:
```
chenzl25:
```suggestion
For example, for sources `s1` and `s2` used in the above section, you can create an interval join:
```

chenzl25 (commented):
I think we can use a better example to illustrate window join and interval join.

docs/sql/query-syntax/query-syntax-join-clause.md
chenzl25: The watermarks of these two sources don't need to be the same, but typically they should use the same windowing table-valued functions (e.g. `TUMBLE`). Watermark columns are used to clean the state after the window is closed.

PR #1058, opened 2023-07-17
add current_setting function

### LINKED ISSUE #1036: Document current_setting()
### Related code PR

https://github.com/risingwavelabs/risingwave/pull/10269

Add a new page: System Administration Functions and document this function.

### Reference

https://www.postgresql.org/docs/current/functions-admin.html

hengm3467:
## Info

- **Description**

Add the system administration function current_setting(), and add several supported settings. Some settings don't have values yet, like `datestyle`, which I didn't include. I'm not sure if we are ready to describe some other settings like `STREAMING_PARALLELISM`. @yezizip2012 What do you think? 

- **Notes**

- **Related code PR**

- **Related doc issue**
  
  Resolves #1036

## For reviewers

- **Preview**

- **Key points**

## Before merging


yezizp2012:
Yes, we don't have a default value for `datastyle`, so it's acceptable for me to not include it for now. 
And for `STREAMING_PARALLELISM`, I think it's okay to include it but might need some explanation for it: if `STREAMING_PARALLELISM` is non-zero, all commands in this session of creating streaming jobs will use it as streaming parallelism.

hengm3467:
I see. Shall we review all parameters and see which ones to describe? cc @neverchanje

image

neverchanje:
I agree that we need descriptions somewhere for these settings. But in order to reduce our efforts, I suggest documenting them through `pg_settings` https://www.postgresql.org/docs/current/view-pg-settings.html, which is the system table for all settings. @yezizp2012 There is a column named 'short_desc' that will provide users their meaning.

Screenshot 2023-07-18 at 16 08 03

https://crate.io/docs/crate/reference/en/5.3/config/session.html CrateDB indeeds maintain a doc. But not all databases did the same.

emile-00 (approved):

hengm3467:
Let's decouple the description of messages and the `current_setting()` function. In this doc PR, let's just list the settings. 

@yezizip2012 FYI. I created some descriptions of the settings, but think it might make more sense to include them in https://github.com/risingwavelabs/risingwave/issues/11014 . Here are the descriptions:

|`server_version`|PostgreSQL server version. |
|`server_version_num`| PostgreSQL server version, in the numeric format. Version `8.3.0` will be displayed as `80300` in the numeric format.|
|`timezone`|Timezone for the current instance. |
|`query_mode`| Query mode of RisingWave. Enum values: `auto`, `batch`, `streaming`. The default value is `auto`.|
|`streaming_parallelism`| The number of actors assigned to each streaming job. If it is set to non-zero, all streaming jobs in the current session will use it as the parallelism setting. |
|`batch_parallelism`| The number of actors assigned to each streaming job. If it is set to non-zero, all streaming jobs in the current session will use it as the parallelism setting.|

PR #1074, opened 2023-07-18
Document BYTES row format

### LINKED ISSUE #1028: Document `BYTES` as a new data format
### Related code PR

https://github.com/risingwavelabs/risingwave/pull/10592

Document `BYTES` as a new data format. Currently only Kafka source is tested, confirm if other sources also support BYTES data format. 

### Reference

_No response_

emile-00:
## Info

- **Description**

  Document BYTES row format under CREATE SOURCE and Ingest data from Kafka.

- **Notes**

- **Related code PR**

  https://github.com/risingwavelabs/risingwave/pull/10592

- **Related doc issue**
  
  Resolves https://github.com/risingwavelabs/risingwave-docs/issues/1028

## For reviewers

- **Preview**

- **Key points**

## Before merging


docs/sql/commands/sql-create-source.md
```diff
@@ -271,6 +271,12 @@ FORMAT UPSERT
 ENCODE AVRO
 ```
 
+### Bytes
+
+Currently, we require users to decode the stream data through a certain format. But the user may want no decoding phase. This is a missing capability, which is fundamental, although uncommon to appear in real-world use cases. We only accept table with exactly one field of BYTEA type in ROW FORMAT BYTES. Only append-only type is considered here because BYTEA cannot be the primary key for the table #10577.
+
+If
```
wugouzi: what is this "if"?

emile-00: Sorry about that! Seems like I made an error when creating the pull request. This isn't the intended description. Please take a look at the updated version. Thank you!

docs/sql/commands/sql-create-source.md
```diff
@@ -271,6 +271,15 @@ FORMAT UPSERT
 ENCODE AVRO
 ```
 
+### Bytes
+
+RisingWave allows you to read data streams without decoding the data by using the `BYTES` row format. However, the table or source can have exactly one field of `BYTEA` data and it cannot be the primary key.
```
wugouzi: It can be the primary key.

wugouzi:
Rest LGTM. Thanks

wugouzi (approved):

hengm3467 (approved):

Issue #869, opened 2023-05-13
Document: Window Functions


emile-00:
### Related code PR

https://github.com/risingwavelabs/risingwave/pull/9597
https://github.com/risingwavelabs/risingwave/pull/9622
https://github.com/risingwavelabs/risingwave/pull/9701

### Which part(s) of the docs might be affected or should be updated? And how?

New reference page under `Functions and operators` on Window functions. Cover the `OVER()`  clause and the `EMIT ON WINDOW CLOSE` clause. Document the `lag()` and `lead()` window functions.

### Reference

Overview of window functions and `lag()` and `lead()`: https://www.sqltutorial.org/sql-window-functions/
Also document `EXCLUDE CURRENT ROW` and `EXCLUDE NO OTHERS` support for window functions

hengm3467:
Close as completed in #1077

PR #1077, opened 2023-07-20
document (over) window functions

### LINKED ISSUE #1046: Document `first_value()` and `last_value()`
### Related code PR

https://github.com/risingwavelabs/risingwave/pull/10740

Document `first_value()` and `last_value()` under aggregate functions

### Reference

_No response_

### LINKED ISSUE #1048: Document row_number window function
### Related code PR

https://github.com/risingwavelabs/risingwave/pull/10869

Window function

### Reference

_No response_

### LINKED ISSUE #1144: Document over window functions: `lead`, `lad`
### Related code PR

https://github.com/risingwavelabs/risingwave/pull/10915

Document over window functions: `lead`, `lad`

### Reference

_No response_

hengm3467:
## Info

- **Description**

Add description and syntax for window functions.

- **Notes**

- **Related code PR**

- **Related doc issue**
  
  Resolves #1144 #1046 #1048 #869

## For reviewers

- **Preview**

- **Key points**

## Before merging


CharlieSYH:
Related: https://github.com/risingwavelabs/risingwave-docs/issues/1144

fuyufjh:
@stdrc Please take a look

docs/transform/window-functions.md
```diff
@@ -0,0 +1,126 @@
...
+
+`window_function` can be one of the following:
+
+- Ranking functions – `row_number()`, `rank()`
+- Aggregate-type functions – `sum()`, `min()`, `max()`, `avg()` and `count()`
+- Value functions – `lead()`, `lag()`, `first_value()`, and `last_value()`
+
+:::note
+
+The `PARTITION` clause is required. If you do not want to partition the rows into smaller sets, you can work around by specifying `PARTITION BY 1::int`.
```
stdrc: > The `PARTITION` clause is required.

"The `PARTITION BY` clause is required currently" may be better. We have plan to support OVER clause w/o `PARTITION BY` later.

docs/transform/window-functions.md
```diff
@@ -0,0 +1,126 @@
...
+( PARTITION BY partition_expression 
+[ ORDER BY sort_expression [ ASC | DESC ] [ NULLS { FIRST | LAST } ] [, ...] ]
+[frame_clause])
+
+```
+
+`window_function` can be one of the following:
+
+- Ranking functions – `row_number()`, `rank()`
+- Aggregate-type functions – `sum()`, `min()`, `max()`, `avg()` and `count()`
```
stdrc: IMO "Aggregate functions" is OK.

docs/transform/window-functions.md
```diff
@@ -0,0 +1,126 @@
...
+- Value functions – `lead()`, `lag()`, `first_value()`, and `last_value()`
+
+:::note
+
+The `PARTITION` clause is required. If you do not want to partition the rows into smaller sets, you can work around by specifying `PARTITION BY 1::int`.
+
+:::
+
+When operating in the emit-on-window-close mode for a streaming query, it is necessary to include an ORDER BY clause for each window function. Please ensure that you specify only one column to order by. This column must be a timestamp column that includes watermarks. It's important to note that when using the timestamp column from this streaming query in another streaming query, the watermark information associated with the column is not retained.
```
stdrc: > This column must be a timestamp column that includes watermarks.

Would be better to write:

> This column, generally a timestamp column, must have a watermark definition on it.

docs/transform/window-functions.md
```diff
@@ -0,0 +1,126 @@
...
+
+```
+UNBOUNDED PRECEDING
+offset PRECEDING
+CURRENT ROW
+offset FOLLOWING
+UNBOUNDED FOLLOWING
+```
+
+Where `offset` in a positive integer. If only `frame_start` is specified, `CURRENT ROW` will be used as the end of the window.
```
stdrc: > Where `offset` in a positive integer.

should be:

> Where `offset` is a positive integer.

docs/transform/window-functions.md
```diff
@@ -0,0 +1,126 @@
...
+
+`last_value()` returns the value of the last row in the current window frame.
+
+The syntax of `last_value()` is:
+
+```sql
+last_value ( value anyelement ) → anyelement
+```
+
+## Aggregate-type window functions
```
stdrc: Similar to before, "aggregate window function" is OK.

docs/transform/window-functions.md
```diff
@@ -0,0 +1,126 @@
...
+[ ORDER BY sort_expression [ ASC | DESC ] [ NULLS { FIRST | LAST } ] [, ...] ]
+[frame_clause])
+
+```
+
+`window_function` can be one of the following:
+
+- Ranking functions – `row_number()`, `rank()`
+- Aggregate-type functions – `sum()`, `min()`, `max()`, `avg()` and `count()`
+- Value functions – `lead()`, `lag()`, `first_value()`, and `last_value()`
```
stdrc: Since we categorize these functions into 2 cats below: general-purpose and aggregate, here we'd better also list 2 categories. Maybe we can list ranking and value functions as 2 sub-categories of general-purpose window functions.

Also, since we list all window functions supported later in this page, we may add something like ", etc" here and give links to the sections below.

docs/transform/window-functions.md
```diff
@@ -0,0 +1,126 @@
...
+
+### `row_number()`
+
+The `row_number()` function assigns a unique sequential integer to each row within a partition of a result set. The numbering starts at 1 for the first row in each partition and increments by 1 for each subsequent row.
+
+`row_number()` can be used to turn non-unique rows into unique rows. This could be used to eliminate duplicate rows.
+
+### `rank()`
+
+Returns the rank of the current row, with gaps; that is, the `row_number` of the first row in its peer group. Only top-N pattern is supported. For details about this pattern, see [Top-N by group](../syntax/sql-pattern-topn.md).
```
stdrc: For both functions, we recommend users to use TopN pattern instead of pure OverWindow. It's a strong recommendation. Maybe better to have a `note` block.

Also, better to have function signatures here, although there's no argument, the return type may matter.

docs/transform/window-functions.md
```diff
@@ -0,0 +1,126 @@
...
+
+The syntax of `last_value()` is:
+
+```sql
+last_value ( value anyelement ) → anyelement
+```
+
+## Aggregate-type window functions
+
+The aggregate-type window functions include `sum()`, `min()`, `max()`, `avg()` and `count()`.
```
stdrc: May give a link to the "Aggregate functions" page. All aggregate functions are supported here, not only the listed 5 ones.

docs/transform/window-functions.md
```diff
@@ -0,0 +1,126 @@
...
+
+Although the calculations that are performed in aggregate-type window functions are similar to those performed by regular aggregate functions, they are different in nature.
+
+Unlike regular aggregate functions, aggregate-type window functions does not group multiple rows into a single output row — the rows retain their separate identities. Behind the scenes, aggregate-type window functions can access more than just the current row of the query result.
```
stdrc: IMHO these 2 paragraphs may be misleading. Aggregate window functions do have the same calculations as regular aggregate function calls. The difference is that for window function call, the aggregation is done for each window frame, so each input row has one corresponding output row. I think these nature can be put at the beginning of this page, cuz it's the nature of all window functions.

docs/transform/window-functions.md
hengm3467: Good point! Let's point to that page for all aggregate functions.

docs/transform/window-functions.md
hengm3467: I see. These two paragraphs are some of my summaries before, and may not be appropriate here. Let me take those out.

docs/transform/window-functions.md
hengm3467: @stdrc I updated the topic based on your comments. Please take a look again. Thanks.

docs/transform/window-functions.md
```diff
@@ -0,0 +1,145 @@
...
+
+The syntax of `rank()` is:
+
+```sql
+rank() → integer
+```
+
+:::note
+
+We recommend using `rank()` only for top-N pattern queries. For details about this pattern, see [Top-N by group](/sql/syntax/sql-pattern-topn.md).
```
stdrc: Sorry I didn't make it clear in last review, `rank()` is now only supported for TopN pattern.

stdrc (approved):
Rest LGTM!

PR #1194, opened 2023-08-22
Doc for iceberg sink v2

liurenjie1024:
## Info

- **Description**

More docs for iceberg sink.

- **Notes**

No.

- **Related code PR**

- **Related doc issue**
  
## For reviewers

- **Preview**

- **Key points**

## Before merging

### DIFF
--- a/docs/guides/sink-to-iceberg.md
+++ b/docs/guides/sink-to-iceberg.md
@@ -11,6 +11,10 @@ This guide describes how to sink data from RisingWave to Apache Iceberg using th
 The Iceberg sink connector in RisingWave is currently an experimental feature. Its functionality is subject to change. We cannot guarantee its continued support in future releases, and it may be discontinued without notice. You may use this feature at your own risk.
 :::
 
+:::caution Breaking change in v1.2
+In version 1.2, we have made significant improvements to the implementation of the feature, which has resulted in changes to the parameters. As a result, applications that rely on the previous version of the feature (specifically, the version included in RisingWave v1.0.0 and v1.1) may no longer function correctly. To restore functionality to your applications, please carefully review the syntax and parameters outlined on this page and make any necessary revisions to your code.
+:::
+
 ## Prerequisites
 
 - Ensure you already have an Iceberg table that you can sink data to.
@@ -44,7 +48,7 @@ WITH (
 
 ## Data Type Mapping
 
-Risingwave converts risingwave data types from/to iceberg according to the following data type mapping table:
+Risingwave converts risingwave data types from/to Iceberg according to the following data type mapping table:
 
 |Risingwave Type| Iceberg Type|
 |---------------|-------------|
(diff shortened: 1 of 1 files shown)


docs/guides/sink-to-iceberg.md
```diff
@@ -24,7 +24,7 @@ The Iceberg sink connector in RisingWave is currently an experimental feature. I
 CREATE SINK [ IF NOT EXISTS ] sink_name
 [FROM sink_from | AS select_query]
 WITH (
-   connector='iceberg',
+   connector='iceberg_v2',
```
wenym1: Since it's currently `v2`, shall we write a new doc instead of in-place modifying the java one?

Later after we stabilize the rust implementation, we can then change the `connector` to `iceberg` in the new doc and deprecate the old one.

liurenjie1024: I've submitted a pr to rename it: https://github.com/risingwavelabs/risingwave/pull/11824. I'm not sure if we should have two versions on the doc for users.

wenym1: Before the rust version fully supports upsert and partition I think we can have two versions of doc. 

After the java version is deprecated we can then remove the doc for the java version.

liurenjie1024: cc @neverchanje Any comments from pm's view?

neverchanje: Giving that the iceberg sink is currently in alpha period, I think users have been aware that this feature is subject to changes. Therefore, I think we can do this in-place change.

neverchanje: However, it would be better if you can leave a warning to remind users that there is a breaking change introduced in the upcoming 1.2 release.  We can remove the warning in 1.3.

liurenjie1024: The second question is should we keep another doc about 'iceberg_java' ?

neverchanje: No. Just keep it simple.

docs/guides/sink-to-iceberg.md
```diff
@@ -33,18 +33,34 @@ WITH (
...
 | s3.secret.key   | Required. Secret key of the S3 compatible object store.|
 | database.name   | Required. The database of the target Iceberg table.|
 | table.name      | Required. The name of the target Iceberg table.|
 
-:::note
-Iceberg sinks with `upsert` type is slower than `append-only`.
-:::
+## Data Type Mapping
+
+Risingwave converts risingwave data types from/to iceberg according to following data type matching:
```
hengm3467: -> according to the following data type mapping table:

hengm3467: "iceberg" -> "Iceberg"

hengm3467:
I made some wording updates, and added a breaking change note. Now it's good to go from my side. Thanks a lot for putting this together. @liurenjie1024

hengm3467 (approved):

liurenjie1024 (commented):
Thanks

PR #1274, opened 2023-09-07
fix: Caused by an incorrect hreflang link

TingJianZheng:
The issue is not conducive to SEO.
image

image

fixed:
image


chrispaul-03 (approved):
LGTM!

PR #1312, opened 2023-09-18
Document the rule about generated columns (cannot be specified as primary keys)

### LINKED ISSUE #1285: Document: fix: ban pk comprising generated column
This issue tracks the documentation update needed for the merged PR #12181.

Source PR URL: https://github.com/risingwavelabs/risingwave/pull/12181
Source PR Merged At: 2023-09-11T07:42:02Z

hengm3467:
## Info

- **Description**

Document the rule that generated columns cannot be specified as primary keys.

- **Notes**

- **Related code PR**

https://github.com/risingwavelabs/risingwave/pull/12181

- **Related doc issue**
  
  Resolves #1285

## For reviewers

- **Preview**

- **Key points**

## Before merging


docs/sql/commands/sql-create-source.md
```diff
@@ -26,23 +28,13 @@ CREATE SOURCE [ IF NOT EXISTS ] source_name (
...
-...
-[WATERMARK FOR col_name as generation_expression]
-```
-
-:::note
-To know when a data record is loaded to RisingWave, you can define a column that is generated based on the processing time (`<column_name> timestampz AS proctime()`) when creating the table or source.
-:::
-
-:::note
+For sources with primary key constraints, if you insert a new data record with an existing key, the new record will overwrite the existing record. A [generated column](/sql/query-syntax/query-syntax-generated-columns.md) cannot be defined as a primary key.
```
BugenZhao: > A [generated column](/sql/query-syntax/query-syntax-generated-columns.md) cannot be defined as a primary key.

I guess this is not true? We only disallow generated columns with impure expressions as the primary key. cc @yuhao-su 

By the way, the error message here doesn't mention `impure` as well, which sounds misleading.

https://github.com/risingwavelabs/risingwave/blob/14292d783896ed01a569f3f3963aec944a1f42ce/src/frontend/src/handler/create_table.rs#L220-L225

yuhao-su: > I guess this is not true?

You are right.
I corrected the statement in my PR description but forgot to notify the doc team. 

The correct statement should be:
> The generated columns with impure expression should not be part of the defined primary key in a table.



>  the error message here doesn't mention impure as well

My bad. Will fix.

hengm3467: @yuhao-su @BugenZhao I feel "impure expression" is a term that needs explanation itself, so I rephrase the description to be like this:

```A generated column that is defined with non-deterministic functions cannot be specified as part of the primary key. For example, if `A1` is defined as `current_timestamp()`, then it cannot be part of the primary key.```

Let me know if this is accurate. If not, please explain to me what impure expressions refer to (with examples). Thanks!

BugenZhao: I think it's good enough. 👍 Actually this is not the only case that purity of the expression matters. Another occurrence is the functional index (named as [_Indexes on Expressions_](https://docs.risingwave.com/docs/current/sql-create-index/#indexes-on-expressions) in the doc), where impure expressions are also disallowed.

https://github.com/risingwavelabs/risingwave/blob/5358f1ab393797663f9b238acfdad4b973621981/src/frontend/src/handler/create_index.rs#L93-L99

I guess we can introduce a separate topic to discuss the purity of expressions in RisingWave in the future.

emile-00 (approved):

docs/sql/commands/sql-create-source.md
hengm3467: Tracked here: https://www.notion.so/risingwave-labs/Pure-and-impure-expressions-0fcb3163649a46f5811aa92b60468cdd?pvs=4

Will come to it when other higher priorities are processed.

PR #1471, opened 2023-11-02
Use a better way to present the data casting table

### LINKED ISSUE #342: Use a better way to present the data casting table
### Related issue

https://github.com/risingwavelabs/risingwave-docs/issues/305

https://www.risingwave.dev/docs/latest/sql-data-types/#casts

and also this https://www.risingwave.dev/docs/latest/sql-function-logical/

### Reference

_No response_

CharlieSYH:
## Info

- **Description**

  - Not that ideal, huh?

image

- **Related doc issue**
  
  Resolves https://github.com/risingwavelabs/risingwave-docs/issues/342

## For reviewers

- **Preview**

- **Key points**

## Before merging


hengm3467:
@CharlieSYH I think it's already much better than the original one. One additional improvement I can think of is to change the  column names:
Implicit -> Implictly cast to
Explicit -> Explictly cast to
Assignment -> Assigned to

And add a note about these three column names (using the original definition is fine).

CharlieSYH:
> @CharlieSYH I think it's already much better than the original one. One additional improvement I can think of is to change the column names: Implicit -> Implictly cast to Explicit -> Explictly cast to Assignment -> Assigned to
> 
> And add a note about these three column names (using the original definition is fine).

@hengm3467 Updated. Please review it again.

hengm3467 (approved):
LGTM. Thanks.

PR #1487, opened 2023-11-07
Doc FLUSH and implicit flush

### LINKED ISSUE #1451: document the flush command
### Related code PR

_No response_

It's been 2yrs since we introduced this command. Generally, we need to let users know that they need to call FLUSH every time they write data. Therefore, it requires both the sql command document and some updates to the INSERT/DELETE/UPDATE doc. 

As I discussed with @hzxa21 offline, I wonder if we can leverage partial checkpoint to enable more frequent flushes from the client, for example, flush on every insert command. As a result, it will allow us to get rid of the FLUSH command. cc @fuyufjh I'll bring up a technical discussion at the kernel side later.

### Reference

_No response_

CharlieSYH:
## Info

- **Description**

  - Add a page for FLUSH
  - Add a page for SET RW_IMPLICIT_FLUSH
  - Add a note in INSERT, UPDATE, & DELETE

- **Related doc issue**
  
  Resolves https://github.com/risingwavelabs/risingwave-docs/issues/1451

## For reviewers

- **Preview**

- **Key points**
@neverchanje 
  - Please verify the accuracy and relevance of the information since the related PRs I found are very old, and there might be updates since then.
  - I also added a page for SET RW_IMPLICIT_FLUSH. Not sure if we want that. If so, please provide a comparison between FLUSH and this configuration option.

## Before merging


hengm3467 (approved):
LGTM. Thanks!

PR #1478, opened 2023-11-06
Doc to_jsonb()

CharlieSYH:
## Info

- **Related code PR**

  - https://github.com/risingwavelabs/risingwave/pull/13161

- **Related doc issue**
  
  Resolves https://github.com/risingwavelabs/risingwave-docs/issues/1475

## For reviewers

- **Preview**

  - [ Paste the preview link to the updated page(s) here. Edit this item after the preview site is ready. To find the updated pages, scroll down to locate and open the Amplify preview link and select the **dev** version of the documentation. ]

- **Key points**

  - [ Parts that may need revision or extra consideration. ]

## Before merging

- [ ] I have checked the doc site preview, and the updated parts look good.

- [ ] I have acquired the approval from the owner (and optionally the reviewers) of the code PR and at least one tech writer (`CharlieSYH`, `emile-00`, & `hengm3467`).

PR #1497, opened 2023-11-09
Revert "Doc to_jsonb()"

CharlieSYH:
Reverts risingwavelabs/risingwave-docs#1478


KeXiangWang (approved):
Sorry. This should not be included in 1.4, so we revert it temporarily.

PR #1526, opened 2023-11-15
Update `docker-compose` to `docker compose` and add notes

### LINKED ISSUE #1504: Update docker compose commands

We received feedback from a user who ran into an error when using `docker compose up -d` to start a cluster but succeeded with `docker-compose up -d`. It seems like the user installed Compose V1.

`docker-compose` is still enabled by default to use Compose V2, but `docker compose` is the recommended syntax after V2. See https://docs.docker.com/compose/migrate/#what-are-the-differences-between-compose-v1-and-compose-v2

So, we may want to:

- Update all `docker-compose` to `docker compose`.
- Add code comments or notes about using Compose V2 or Docker Desktop.

CharlieSYH:
## Info

- **Description**

  - See https://github.com/risingwavelabs/risingwave-docs/issues/1504

- **Related doc issue**
  
  Resolves https://github.com/risingwavelabs/risingwave-docs/issues/1504

## For reviewers

- **Preview**

- **Key points**

## Before merging

### DIFF
--- a/docs/deploy/risingwave-docker-compose.md
+++ b/docs/deploy/risingwave-docker-compose.md
@@ -65,6 +65,12 @@ This is the default option. To start a RisingWave cluster with MinIO as the stor
 docker compose up -d
 ```
 
+:::tip Command not found?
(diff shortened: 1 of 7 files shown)


emile-00 (approved):

neverchanje (approved):
Thanks. LGTM

hengm3467 (approved):
LGTM. Thanks!

PR #1541, opened 2023-11-19
Create query-syntax-owner-to-clause.md

### LINKED ISSUE #1490: Document: feat: support `ALTER DATABASE/SCHEMA/TABLE/... xxx OWNER TO xxx` syntax
This issue tracks the documentation update needed for the merged PR #13216.

Source PR URL: https://github.com/risingwavelabs/risingwave/pull/13216
Source PR Merged At: 2023-11-08T03:16:34Z

ShanlanLi:
## Info

- **Description**

  - [ Add OWNER TO clause.]

- **Notes**

- **Related code PR**

  - [ https://github.com/risingwavelabs/risingwave/pull/13216 ]

- **Related doc issue**
  
  Resolves [ https://github.com/risingwavelabs/risingwave-docs/issues/1490 ]

## For reviewers

- **Preview**

- **Key points**

## Before merging

### DIFF
--- a/docs/sql/query-syntax/query-syntax-owner-to-clause.md
+++ b/docs/sql/query-syntax/query-syntax-owner-to-clause.md
@@ -7,7 +7,7 @@ title: OWNER TO clause
   <link rel="canonical" href="https://docs.risingwave.com/docs/current/query-syntax-owner-to-clause/" />
 </head>
 
-"OWNER TO" is a clause that is used to assign or change the owner of an object, including the database, schema, table, materialized view, source, and sink. It is typically used in conjunction with the ALTER statement.
+`OWNER TO` is a clause that is used to assign or change the owner of an object, including the database, schema, table, materialized view, source, and sink. It is typically used in conjunction with the `ALTER` statement.
 
 Here is the basic syntax of an `ALTER` statement with the `OWNER TO` clause:
 
--- a/docs/sql/query-syntax/query-syntax-owner-to-clause.md
+++ b/docs/sql/query-syntax/query-syntax-owner-to-clause.md
@@ -1,0 +1,70 @@
+---
+id: query-syntax-owner-to-clause
+slug: /query-syntax-owner-to-clause
+title: OWNER TO clause
+---
+<head>
+  <link rel="canonical" href="https://docs.risingwave.com/docs/current/query-syntax-owner-to-clause/" />
+</head>
+
+"OWNER TO" is a clause that is used to assign or change the owner of an object, including the database, schema, table, materialized view, source, and sink. It is typically used in conjunction with the ALTER statement.
+
+Here is the basic syntax of an `ALTER` statement with the `OWNER TO` clause:
+
+```sql
+ALTER { DATABASE | SCHEMA | TABLE | [MATERIALIZED] VIEW | SOURCE | SINK } target OWNER TO new_user
+```
+Basic `OWNER TO` clause examples:
+
+
+
+```sql title=Example
+-- Change the owner of a database
+ALTER DATABASE d OWNER TO user1;
+```
+ 
+
+```sql title=Example
+-- Change the owner of a schema
+ALTER SCHEMA s OWNER TO user1;
+```
+
+
+
+```sql title=Example
+-- Change the owner of a table
+ALTER TABLE t OWNER TO user1;
+```
+
+
+
+
+```sql title=Example
+-- Change the owner of a view
+ALTER VIEW v OWNER TO user1;
+```
+
+
+
+
+```sql title=Example
+-- Change the owner of a materialized view
+ALTER MATERIALIZED VIEW mv OWNER TO user1;
+```
+
+
+
+
+```sql title=Example
+-- Change the owner of a source
+ALTER SOURCE src OWNER TO user1;
+```
+
+
+
+
+```sql title=Example
+-- Change the owner of a sink 
+ALTER SINK sink OWNER TO user1;
+```
+
--- a/docs/sql/query-syntax/query-syntax-owner-to-clause.md
+++ b/docs/sql/query-syntax/query-syntax-owner-to-clause.md
@@ -15,7 +15,7 @@ Here is the basic syntax of an `ALTER` statement with the `OWNER TO` clause:
 ALTER { DATABASE | SCHEMA | TABLE | [MATERIALIZED] VIEW | SOURCE | SINK } target OWNER TO new_user
 ```
 :::info
-Note that the statement `ALTER TABLE target OWNER TO new_user` will also change the owner of the table's indexes, internal tables and the associated source (if the table is created using `CREATE TABLE table_name WITH (connector='connector_name')`).
+Note that the statement `ALTER ... OWNER TO ...` will cascadingly change all related sub-objects as well.
 :::
 
 Basic `OWNER TO` clause examples:
--- a/docs/sql/query-syntax/query-syntax-owner-to-clause.md
+++ b/docs/sql/query-syntax/query-syntax-owner-to-clause.md
@@ -14,6 +14,10 @@ Here is the basic syntax of an `ALTER` statement with the `OWNER TO` clause:
 ```sql
 ALTER { DATABASE | SCHEMA | TABLE | [MATERIALIZED] VIEW | SOURCE | SINK } target OWNER TO new_user
 ```
+:::info
+Note that the satement `ALTER TABLE target OWNER TO new_user` will also change the owner of the table's indexes, internal tables and the associated source (if the table is created using `CREATE TABLE table_name WITH (connector='connector_name')`).
+:::
+
 Basic `OWNER TO` clause examples:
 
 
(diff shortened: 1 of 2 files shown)


ShanlanLi:
Problems to ask for suggestions from the engineer and writer before asking for review:
+ Whether the category "query syntax" is proper.
+ We have ALTER SOURCE, ALTER TABLE, ALTER USER in COMMANDS now, but don't have ALTER DATABASE/SCHEMA/SINK. Do I need to add them aside from OWNER TO?

Rossil2012:
I think it's worth mentioning that:
`ALTER TABLE target OWNER TO new_user`: This statement will also change the owner of the table's indexes, internal tables and the associated source (if the table is created using `CREATE TABLE t WITH (connector='xxx')`).

docs/sql/query-syntax/query-syntax-owner-to-clause.md
```diff
@@ -0,0 +1,70 @@
...
+-- Change the owner of a schema
+ALTER SCHEMA s OWNER TO user1;
+```
+
+
+
+```sql title=Example
+-- Change the owner of a table
+ALTER TABLE t OWNER TO user1;
+```
```
Rossil2012, line 41: I think it's worth mentioning that:
ALTER TABLE target OWNER TO new_user: This statement will also change the owner of the table's indexes, internal tables and the associated source (if the table is created using CREATE TABLE t WITH (connector='xxx')).
FYI, see https://github.com/risingwavelabs/risingwave/pull/13419.

docs/sql/query-syntax/query-syntax-owner-to-clause.md
```diff
@@ -0,0 +1,74 @@
+---
+id: query-syntax-owner-to-clause
+slug: /query-syntax-owner-to-clause
+title: OWNER TO clause
+---
+<head>
+  <link rel="canonical" href="https://docs.risingwave.com/docs/current/query-syntax-owner-to-clause/" />
+</head>
+
+"OWNER TO" is a clause that is used to assign or change the owner of an object, including the database, schema, table, materialized view, source, and sink. It is typically used in conjunction with the ALTER statement.
```
CharlieSYH: We usually use code format for commands and clauses in our docs: `OWNER TO` and `ALTER`

CharlieSYH (approved):
LGTM

yezizp2012 (approved):
LGTM

Rossil2012 (approved):
LGTM.

docs/sql/query-syntax/query-syntax-owner-to-clause.md
```diff
@@ -0,0 +1,74 @@
...
+
+`OWNER TO` is a clause that is used to assign or change the owner of an object, including the database, schema, table, materialized view, source, and sink. It is typically used in conjunction with the `ALTER` statement.
+
+Here is the basic syntax of an `ALTER` statement with the `OWNER TO` clause:
+
+```sql
+ALTER { DATABASE | SCHEMA | TABLE | [MATERIALIZED] VIEW | SOURCE | SINK } target OWNER TO new_user
+```
+:::info
+Note that the statement `ALTER TABLE target OWNER TO new_user` will also change the owner of the table's indexes, internal tables and the associated source (if the table is created using `CREATE TABLE table_name WITH (connector='connector_name')`).
```
neverchanje: Hi, @ShanlanLi 

A table doesn't have so-called 'internal tables'. Only materialized view has. A more accurate saying is that 'ALTER ... OWNER TO ... statement will cascadingly change all related sub-objects as well.'  

Let me check the behavior in details. BTW, thanks @CharlieSYH for bringing this up.

ShanlanLi: Thanks! I've changed this part to:  
:::info
Note that the statement `ALTER ... OWNER TO ...` will cascadingly change all related sub-objects as well.
:::

neverchanje: I've doubled-checked the behavior. Indeed, internal tables are now accessible by the new owner. But the behavior on indexes seems unexpected. 

When the original owner, say, `root`, creates an index on the table, the new owner will still be inaccessible to the index via `select * from idx;`. In the other words: a owner to table doesn't have the access to indexes created by other users. @yezizp2012 Let's discuss the behavior later.

For the word "sub-objects", I found it may not be accurate as tables under a database will be unaffected when the database's owner changes. For example, A is the owner for a database as well as all the tables created in the database. When I change the database owner to B, the tables' owner will remain to be A rather than B. Likewise for schemas.

The best wording might be:

> Note that the statement ALTER [MATERIALIZED VIEW | SINK] ... OWNER TO ... will accordingly change the visibility of all related internal tables as well.

Tables do not have internal tables.

Rossil2012: >When the original owner, say, root, creates an index on the table, the owner is unables to access the index via select * from idx;. In short: a owner to table doesn't have the access to the index created by other users. 

Trying to create an index on a table without ownership and superuser in Postgres:
```
postgres=> create index test_index on test_table(u);
ERROR:  must be owner of table test_table
```
Only the table owner and superuser can create index. So we should either omit the owner property of an index (and check the privileges of its table) or change its owner together with its table (and check the privileges of itself). Nonetheless, maybe we should create an issue of `create index`?

>For the word "sub-objects", I found it may not be accurate as tables under a database will still unchange when the database's owner changes. For example, A is the owner for a database as well as all the tables created in the database. When I change the owner to B, the tables' owner remains to be A instead of B. Likewise for schemas.

It is expected.

>The best wording might be:
Note that the statement ALTER [MATERIALIZED VIEW | SINK] ... OWNER TO ... will accordingly change the visibility of all related internal tables as well.
Tables do not have internal tables.

For now, `ALTER SINK` will change the owner of its internal table. And as `MATERIALIZED VIEW` and `TABLE` are equivalently treated as `table` in `Frontend` and `Meta`, owner of the `table`'s associated source, internal tables and indexes will be all changed correspondingly. To be more precise, for now, owner of `MATERIALIZED VIEW`'s internal tables and indexes, `TABLE`'s associated source and indexes will be changed.

neverchanje: > Only the table owner and superuser can create index. So we should either omit the owner property of an index (and check the privileges of its table) or change its owner together with its table (and check the privileges of itself). Nonetheless, maybe we should create an issue of create index?

The `root`, a superuser, can create an index for the table owner `A`, who is not a superuser. In this case, `A` will not be allowed to read the index. I also doubt if we should allow everyone access the index, because it may contain privacy data. I think we should modify privileges accordingly to let at least owners access.

Please help create an issue and confirm it's feasible.

yezizp2012: Good catch, I prefer to: regardless of whether it is a super user or an owner creating an index, the owner of the table will be set as the owner of the index, its index_table, and internal tables. This approach minimizes changes and does not involve modifications to the privilege verification section, which is also reasonable.

docs/sql/query-syntax/query-syntax-owner-to-clause.md
```diff
@@ -0,0 +1,74 @@
...
+
+`OWNER TO` is a clause that is used to assign or change the owner of an object, including the database, schema, table, materialized view, source, and sink. It is typically used in conjunction with the `ALTER` statement.
+
+Here is the basic syntax of an `ALTER` statement with the `OWNER TO` clause:
+
+```sql
+ALTER { DATABASE | SCHEMA | TABLE | [MATERIALIZED] VIEW | SOURCE | SINK } target OWNER TO new_user
+```
+:::info
+Note that the statement `ALTER ... OWNER TO ...` will cascadingly change all related sub-objects as well.
```
yezizp2012, line 18:
```suggestion
Note that for `TABLE`, `INDEX`, `MATERIALIZED VIEW` and `SINK` the statement `ALTER ... OWNER TO ...` will cascadingly change all related internal-objects as well, and altering on `TABLE` will also change its associated indexes.
```

yezizp2012: Cc @neverchanje WDYT? I think we can get this PR merged firstly, and let the bug about the index be fixed later.

docs/sql/query-syntax/query-syntax-owner-to-clause.md
```diff
@@ -0,0 +1,74 @@
+---
```
hengm3467, line 1: Sorry for not bringing up it earlier, but I think `OWNER TO` is not worth a dedicated topic. It is definitely not on the same level as WITH/FROM/GROUP BY clauses, which can be applied to almost any queries. 

It makes more sense if we update the different ALTER ... commands to include `OWNER TO` as a new `alter_action`. 

We have list out all supported alter_action values in these topics. `ALTER TABLE` can be used as a reference:
https://docs.risingwave.com/docs/current/sql-alter-table/

The way we structure `ALTER SOURCE` should also be changed. 

The alternative is to have a dedicated topic by named it `ALTER ... OWNER TO`. See this link as example:
https://materialize.com/docs/sql/alter-owner/

ShanlanLi:
This doc issue has been solved here:  https://github.com/risingwavelabs/risingwave-docs/pull/1562#event-11100006683. We add these clauses to each alter XXX command now.

PR #1544, opened 2023-11-20
feat(docs): Ingest from kafka

CAJan93:
## Info

- **Description**

- Quick description telling users how they can get infos about the internal columns of a kafka source. 

## For reviewers

- **Preview**

- **Key points**

## Before merging

### DIFF
--- a/versioned_docs/version-1.4/ingest/ingest-from-kafka.md
+++ b/versioned_docs/version-1.4/ingest/ingest-from-kafka.md
@@ -328,6 +328,7 @@ describe source_name;
  _row_id             | serial                      | true      |
 ```
 
+You can select other internal columns, just like you selected `_rw_kafka_timestamp`.
 
 ## Read schemas from locations
 
--- a/versioned_docs/version-1.4/ingest/ingest-from-kafka.md
+++ b/versioned_docs/version-1.4/ingest/ingest-from-kafka.md
@@ -316,6 +316,19 @@ SELECT * FROM source_name
 WHERE _rw_kafka_timestamp > now() - interval '10 minute';
 ```
 
+## Other internal columns
+
+When ingesting from Kafka, RisingWave will create more internal columns. To see them, please use 
+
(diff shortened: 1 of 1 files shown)


fuyufjh:
Hmmm, this is misleading...

`_rw_kafka_timestamp` was already documented at https://docs.risingwave.com/docs/dev/ingest-from-kafka/#query-kafka-timestamp

`_row_id` is not only for Kafka but for all kinds of Sources. We didn't document it because we thought the columns should never be used by users directly. It's okay to document it, but I didn't find a good place for it.

CAJan93:
> It's okay to document it, but I didn't find a good place for it

Do you have a suggestion where to document it? How about here https://github.com/risingwavelabs/risingwave-docs/blob/main/docs/sql/commands/sql-describe.md ?

hengm3467:
@CAJan93 I'm curious about when `_row_id` will be used. If there is some trick that is applicable to all sources, then we can document this in the overview of ingesting data.

CAJan93:
> I'm curious about when _row_id will be used. 

This is a question for @st1page 😁 I have not looked at the code, but AFAIK `_row_id` is generated by RW not Kafka and it is the primary ID by default, so I guess it is just a counter?

st1page:
> > I'm curious about when _row_id will be used.
> 
> This is a question for @st1page 😁 I have not looked at the code, but AFAIK `_row_id` is generated by RW not Kafka and it is the primary ID by default, so I guess it is just a counter?

For batch query, `SELECT _row_id FROM source` will generated a NULL value.
For Streaming query, such as `CREATE MATERIALIZED VIEW AS SELECT _row_id FROM source`, RW will add a row id generator  after the source data's ingestion with a snowflake-like algorithm(https://en.wikipedia.org/wiki/Snowflake_ID) which can assign a unique id for each record.

fuyufjh:
These part has been heavily refactored. Continue on #1750

PR #1609, opened 2023-12-06
Adding user-requested information to the S3 source doc

### LINKED ISSUE #1554: Bug: S3 docs missing key information
### Where's the bug?

https://docs.risingwave.com/docs/0.1.16/create-source-s3

### Describe the bug

This page left me with more questions than answers:

- There is no prefix argument, so does this mean the source reads in all the objects in that S3 bucket?
- Do the objects have to be csv files?  If not, what are my other options?
- Can I create a column with the source file name?
- If new files show up in the bucket do they automatically get slurped in?
- Does this happen on some sort of configurable cadence or is it triggered somehow or is it just the once when I run the command?
- What happens when one file in the bucket is something unexpected (an image or something)?  Does it process everything except that file?  Does it fail completely?
- What happens when part of a file is poorly formatted?

CharlieSYH:
## Info

- **Related doc issue**
  
  Resolves https://github.com/risingwavelabs/risingwave-docs/issues/1554

## For reviewers

- **Preview**

- **Key points**

## Before merging

### DIFF
--- a/docs/ingest/ingest-from-s3.md
+++ b/docs/ingest/ingest-from-s3.md
@@ -8,7 +8,7 @@ slug: /ingest-from-s3
   <link rel="canonical" href="https://docs.risingwave.com/docs/current/ingest-from-s3/" />
 </head>
 
-Use the SQL statement below to connect RisingWave to an Amazon S3 source.
+Use the SQL statement below to connect RisingWave to an Amazon S3 source. RisingWave supports both CSV and [ndjson](https://ndjson.org/) file formats.
 
 ## Syntax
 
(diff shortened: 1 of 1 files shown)


docs/ingest/ingest-from-s3.md
```diff
@@ -184,3 +184,31 @@ WITH (
...
+
+RisingWave has a prefix argument designed for filtering objects in the S3 bucket. It relies on [Apache Opendal](https://github.com/apache/incubator-opendal) whose prefix filter implementation is expected to be released soon.
+
+### Source file name as column
+
+A feature to create a column with the source file name is currently under development. You can track the progress [here](https://github.com/risingwavelabs/rfcs/pull/79).
+
+### Handling new files in the bucket
+
+RisingWave automatically ingests new files added to the bucket. However, it will not detect updates to a file if a file is deleted and a new file with the same name is added at the same time.
```
tabVersion: rw will ignore file delete as well.

docs/ingest/ingest-from-s3.md
```diff
@@ -184,3 +184,31 @@ WITH (
...
+
+### Reading data from the source
+
+You need to create a materialized view from the source or create a table with s3 connector to read the data. Here are some examples:
+
+```sql
+CREATE MATERIALIZED VIEW mv AS SELECT * FROM s3_source;
```
tabVersion, line 209: I think there gonna be a extra one for `create source ... with (connector = 's3_v2')`

tabVersion (approved):

docs/ingest/ingest-from-s3.md
CharlieSYH: Updated.

WanYixian (approved):
LGTM!

hengm3467 (approved):

PR #1613, opened 2023-12-07
Update TiDB data mapping table

### LINKED ISSUE #1575: Document: fix(integration_tests): use `timestamptz` for MySQL, TiDB, Doris and StarRocks
This issue tracks the documentation update needed for the merged PR #13641.

Source PR URL: https://github.com/risingwavelabs/risingwave/pull/13641
Source PR Merged At: 2023-11-29T08:35:43Z

WanYixian:
## Info

- **Description**

  For the data mapping table between RW and TiDB:

   - Switch the positions of two columns: move RW to the left column and TiDB to the right.
   - Add a row for RW `TIMESTAMPTZ` to TiDB `TIMESTAMP`.
   - Add a note explaining how TiDB processes the `TIMESTAMPTZ` data type from RW.

- **Related code PR**

     https://github.com/risingwavelabs/risingwave/pull/13641

- **Related doc issue**
  
  Resolves https://github.com/risingwavelabs/risingwave-docs/issues/1575

## For reviewers

- **Preview**

- **Key points**

## Before merging

### DIFF
--- a/docs/guides/sink-to-tidb.md
+++ b/docs/guides/sink-to-tidb.md
@@ -19,9 +19,8 @@ The following table shows the corresponding data types between RisingWave and Ti
 | RisingWave type | TiDB type |
 |-----------------|------------|
 | BOOLEAN | BOOLEAN |
-| SMALLINT | TINYINT |
-| SMALLINT | SMALLINT |
-| INT | MEDIUMINT |
+| SMALLINT | TINYINT/SMALLINT |
+| INT | INT/MEDIUMINT |
 | BIGINT | BIGINT |
 | REAL | FLOAT |
 | DOUBLE | DOUBLE |
@@ -29,15 +28,7 @@ The following table shows the corresponding data types between RisingWave and Ti
 | DATE | DATE |
 | TIMESTAMP | DATETIME |
 | TIME | TIME |
-| TIMESTAMP | TIMESTAMP |
 | TIMESTAMPTZ | TIMESTAMP |
-| VARCHAR | CHAR |
-| BYTEA | BINARY |
-| BYTEA | VARBINARY |
-| BYTEA | BLOB |
-| TEXT | TEXT |
+| VARCHAR | VARCHAR/CHAR/TEXT/TINYTEXT/<br/>MEDIUMTEXT/LONGTEXT |
+| BYTEA | VARBINARY/BINARY/BLOB/TINYBLOB/<br/>MEDIUMBLOB/LONGBLOB |
 | JSONB | JSON |
-
-:::note
-`TIMESTAMPTZ` types in RisingWave will be converted to `TIMESTAMP` when sinked to TiDB. TiDB doesn't natively support `TIMESTAMPTZ`, but it processes timezone conversions during storage and retrieval. For more information, see [Timezone Handling](https://docs.pingcap.com/tidb/stable/data-type-date-and-time#timezone-handling).
-:::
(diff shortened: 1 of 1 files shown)


docs/guides/sink-to-tidb.md
```diff
@@ -16,23 +16,28 @@ For the syntax, settings, and examples, see [Sink data from RisingWave to MySQL
...
-| TiDB type  | RisingWave type |
-|------------|-----------------|
+| RisingWave type | TiDB type |
+|-----------------|------------|
 | BOOLEAN | BOOLEAN |
-| TINYINT | SMALLINT |
+| SMALLINT | TINYINT |
 | SMALLINT | SMALLINT |
-| MEDIUMINT | INT |
+| INT | MEDIUMINT |
```
xiangjinwu: https://docs.risingwave.com/docs/current/sink-to-mysql-with-jdbc/#data-type-mapping
* `INT` maps to `INT` (4-byte). It may also work for `MEDIUMINT` (3-byte) when no overflow happens.
* Remove `TIMESTAMP | TIMESTAMP`. It is considered misuse (although maybe doable) and `TIMESTAMP` shall always be `DATETIME`.
* `VARCHAR` maps to `VARCHAR`/`CHAR`/`TEXT`/`TINYTEXT`/`MEDIUMTEXT`/`LONGTEXT`. In RisingWave `TEXT` (line 38) is just an alias to `VARCHAR`, just like `DECIMAL` = `NUMERIC` and `INT` = `INTEGER`, and seems we should not have another row for it.
* `BYTEA` maps to `VARBINARY`/`BINARY`/`BLOB`/`TINYBLOB`/`MEDIUMBLOB`/`LONGBLOB`.

docs/guides/sink-to-tidb.md
```diff
@@ -16,23 +16,28 @@ For the syntax, settings, and examples, see [Sink data from RisingWave to MySQL
...
+| BYTEA | BINARY |
+| BYTEA | VARBINARY |
+| BYTEA | BLOB |
 | TEXT | TEXT |
-| JSON | JSONB |
\ No newline at end of file
+| JSONB | JSON |
+
+:::note
+`TIMESTAMPTZ` types in RisingWave will be converted to `TIMESTAMP` when sinked to TiDB. TiDB doesn't natively support `TIMESTAMPTZ`, but it processes timezone conversions during storage and retrieval. For more information, see [Timezone Handling](https://docs.pingcap.com/tidb/stable/data-type-date-and-time#timezone-handling).
```
xiangjinwu: Conceptually,
* `timestamptz` in PG/RW is `timestamp` in MySQL/TiDB. They both store UTC time and do timezone conversions during storage and retrieval.
* `timestamp` in PG/RW is `datetime` in MySQL/TiDB. They both store naive time without timezone.

So it is not whether something is supported or not. It is just different naming.

Instead, we can warn that in **source**, ideally TiDB `timestamp` should become RW `timestamptz`, in the opposite direction of the table above. But when using TiCDC with **canal** json format (rather than **debezium** mysql connector), TiDB `timestamp` would become RW `timestamp`.

docs/guides/sink-to-tidb.md
WanYixian: Revised, pls check.

xiangjinwu (approved):

hengm3467 (approved):
LGTM. Thanks!

PR #1643, opened 2023-12-13
update the monitor-rw-metrics demo doc

neverchanje:
## Info

- **Description**

  Since we have supported a standalone mode, the screenshot should be changed accordingly as well. Because there is no frontend-node-0 anymore. I also found that the results of the demo were changed.

## For reviewers

- **Preview**

- **Key points**

## Before merging


hengm3467 (approved):
LGTM. Thanks.

PR #1663, opened 2023-12-20
Update clickhouse sink type mapping

xuefengze:
## Info

- **Description**

 Update clickhouse sink data type mapping

- **Notes**

- **Related code PR**

https://github.com/risingwavelabs/risingwave/pull/14026

- **Related doc issue**
  
## For reviewers

- **Preview**

- **Key points**

## Before merging

### DIFF
--- a/docs/guides/sink-to-clickhouse.md
+++ b/docs/guides/sink-to-clickhouse.md
@@ -175,7 +175,7 @@ WITH (
 
 |RisingWave Data Type  | ClickHouse Data Type |
 |--------------------- |--------------------- |
-|bool                  | Bool                 |
+|boolean               | Bool                 |
 |smallint              | Int16                |
 |integer               | Int32                |
 |bigint                | Int64                |
--- a/docs/guides/sink-to-clickhouse.md
+++ b/docs/guides/sink-to-clickhouse.md
@@ -2,7 +2,7 @@
 id: sink-to-clickhouse
 title: Sink data from RisingWave to ClickHouse
 description: Sink data from RisingWave to ClickHouse.
-slug: /sink-to-clickhouse 
+slug: /sink-to-clickhouse
 ---
 <head>
   <link rel="canonical" href="https://docs.risingwave.com/docs/current/sink-to-clickhouse/" />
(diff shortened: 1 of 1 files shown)


docs/guides/sink-to-clickhouse.md
```diff
@@ -175,19 +175,23 @@ WITH (
...
+|bigint                | Int64                |
+|real                  | Float32              |
+|double precision      | Float64              |
 |decimal               | Decimal              |
+|character varying     | String               |
+|bytea                 | Not supported        |
 |date                  | Date32               |
-|varchar               | String               |
+|time without time zone| Not supported        |
+|timestamp             | Not supported        |
```
neverchanje, line 189: Thanks for the doc. I am just wondering how do we plan to support these two types as well as `interval`? It seems not appropriate to simply return an error.

neverchanje: Not sure if it because https://clickhouse.com/docs/en/sql-reference/data-types/special-data-types/interval clickhouse's interval type is incompatible with ours. However they do provide the type, so I think there is always a way for the type conversion.

xuefengze: cc @xxhZs

xxhZs: Kind of don't remember, that seems to be the reason, will go do some testing afterward to see if I can implement it

xxhZs (approved):

docs/guides/sink-to-clickhouse.md
xxhZs: Kind of don't remember, that seems to be the reason, will go do some testing afterward to see if I can implement it

docs/guides/sink-to-clickhouse.md
```diff
@@ -175,19 +175,23 @@ WITH (
...
-|bool                  | Bool                 |
-|smallint              | UInt16 or Int16      |
-|int                   | UInt32 or Int32      |
-|bigint                | UInt64 or Int64      |
-|real(float32)         | Float32              |
-|float(float64)        | Float64              |
+|boolean               | Bool                 |
+|smallint              | Int16                |
+|integer               | Int32                |
+|bigint                | Int64                |
```
xxhZs: Currently our code is directly inserting the int in uint format,

WanYixian (approved):
LGTM, thanks!

docs/guides/sink-to-clickhouse.md
xuefengze: I changed it back to `Intxx or UIntxx`.

PR #1695, opened 2023-12-27
Doc use cases and best practices for ingesting data with DML

### LINKED ISSUE #1625: List insert/update/delete as a means of data ingestion
### Related code PR

_No response_

Screenshot 2023-12-08 at 19 10 52

Seems we missed that DML (insert/update/delete) is also data ingestion approach. I think we need a separate page in this section, articulating the best practices for using DML.

### Reference

_No response_

CharlieSYH:
## Info

- **Description**

  - [ What's changed? Which parts of the docs are affected? ]

- **Notes**

- **Related doc issue**
  
  Resolves https://github.com/risingwavelabs/risingwave-docs/issues/1625

## For reviewers

- **Preview**

- **Key points**

## Before merging


neverchanje:
Awesome! Thanks for providing the document. I will see how to refine based on your draft

ShanlanLi (approved):
LGTM

WanYixian (approved):
LGTM

docs/ingest/data-ingestion-dml.md
```diff
@@ -0,0 +1,81 @@
...
+
+  If you need to add additional data to a table, you can use an `INSERT` statement. This is particularly useful for adding small amounts of data without needing to create a new data source. However, for large amounts of data, consider creating a source or declaring `append only` on a table, as discussed in [Best practices for optimizing performance](/performance/performance-best-practices.md).
+
+- **Data transformation**
+
+  You can use DML statements to transform the data in RisingWave. For example, you can use an `UPDATE` statement to change the format of a date column. Remember to create indexes on the columns used in the select statement and the where condition to optimize the performance of these transformations.
+
+- **Testing and development**
+
+  During testing and development, it's often easier to use DML statements to insert test data into RisingWave rather than setting up data sources. However, be mindful of the resources allocated to each query, as discussed in [Performance-related FAQs](/performance/performance-faq.md).
```
neverchanje, line 45: TBH, I feel the information density of these paragraphs is not good enough. I think no user would be unware of INSERT UPDATE DELETE when choosing RisingWave, otherwise they should find general SQL tutorial instead of reading RisingWave documents. Even Cockroach does not provide an overview doc for these operations. They might have assumed that users already knew them.

- Note that updates/deletes are not processed for sources and append-only tables.
- for large amounts of data, consider creating a source or declaring `append only` on a table, as discussed in [Best practices for optimizing performance](/performance/performance-best-practices.md).

From my personal perspective, I only found these points useful. 🤔 👆 The rest of the content seems like what basically every SQL boy should know.

---

>  Remember to create indexes on the columns used in the select statement and the where condition to optimize the performance of these transformations.

I am also a bit unclear why you mentioned CREATE INDEX in this paragraph, which seems irrelevant.

docs/ingest/data-ingestion-dml.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Best practices for using DML for data ingestion
+
+When using DML for data ingestion, consider the following best practices:
+
+- **Use DML for small amounts of data**
+
+  For large amounts of data, consider creating a source or declaring `append only` on a table.
```
neverchanje, line 53: It is repeated.
I recommend referring the suggestion to the 'append-only' page. https://docs.risingwave.com/docs/current/performance-best-practices/#when-to-create-a-source-or-declare-append-only-on-a-table

docs/ingest/data-ingestion-dml.md
```diff
@@ -0,0 +1,81 @@
...
+## Using DML for data ingestion
+
+DML operations can be used for data ingestion in RisingWave in the following ways:
+
+- **`INSERT`**: To add new data to a table.
+
+- **`UPDATE`**: To modify existing data in a table. Note that updates are not processed for sources and append-only tables.
+
+- **`DELETE`**: To remove data from a table. Note that deletes are not processed for sources and append-only tables.
+
```
neverchanje, line 22: You may add links of each operation doc respectively.

https://docs.risingwave.com/docs/current/sql-update/
https://docs.risingwave.com/docs/current/sql-insert/

neverchanje:
There are actually a few things not covered. 

1. FLUSH. We need to let users know that their operations will not be committed immediately unless a 'FLUSH' is invoked. Alternatively, they can https://docs.risingwave.com/docs/current/sql-set-rw-implicit-flush/ `SET RW_IMPLICIT_FLUSH = true` that applies at the session level. 
RisingWave's object-store-based storage discourages frequent flushing per individual record, since it's optimized for small batches. It's hence recommended to ammulate a few changes in memory, e.g, around 100 records, before executing FLUSH. Under the hood, RisingWave will consolidate these changes into one batch and write it into object storage as an object. Most object storage, such as S3, doesn't deliver optimal performance when handling writes on a record-by-record basis.
A common practice for batch ingestion:

```sql
INSERT INTO your_table_name (column1, column2, column3, ...)
VALUES
  (value1_row1, value2_row1, value3_row1, ...),
  (value1_row2, value2_row2, value3_row2, ...),
  (value1_row3, value2_row3, value3_row3, ...),
  ...
  (value1_last_row, value2_last_row, value3_last_row, ...);

FLUSH
```

6. INSERT INTO SELECT 
https://github.com/risingwavelabs/risingwave-docs/issues/1771 I recently wrote a use pattern that leverages this syntax. Perhaps we can link it as well.

PR #1717, opened 2024-01-04
Add note on max_wal_senders for PG CDC

### LINKED ISSUE #1670: Change required: PostgreSQL CDC max_wal_senders limitation
### Related code PR

There is a limitation by PG configuration `max_wal_senders`. Every table requires a sender if we create tables from the pg CDC directly. So the `max_wal_senders` must be greater or equal to the synced table numbers. Refer to https://github.com/risingwavelabs/risingwave/issues/14055

And if we use shared PG CDC source after https://github.com/risingwavelabs/risingwave/pull/13958 merged in the future, we can avoid to increase `max_wal_senders` when using shared source. But we still need to increase it when we create tables from the pg CDC directly.

So our documentation should mention this limitation for PG CDC.

Remind users to increase `max_wal_senders` in https://docs.risingwave.com/docs/current/ingest-from-postgres-cdc/ for releases if they use `create table` from pg CDC instead of shared source.

### Reference

_No response_

emile-00:
## Info

- **Description**

  - Add note on setting max_wal_senders for PG CDC tables

- **Notes**

- **Related code PR**

- **Related doc issue**
  
  Resolves https://github.com/risingwavelabs/risingwave-docs/issues/1670

## For reviewers

- **Preview**

- **Key points**

## Before merging


cyliu0 (commented):
According to the [native PG max_wal_senders doc](https://www.postgresql.org/docs/16/runtime-config-replication.html#GUC-MAX-WAL-SENDERS), `The default is 10`.

According to the [AWS RDS doc](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_PostgreSQL.Replication.ReadReplicas.html), `The default for RDS for PostgreSQL 13 and higher releases is 20`

The `16` you found in our previous issue is defined in our kube-bench PG configuration.

cyliu0 (approved):
LGTM

WanYixian (approved):
LGTM

PR #1960, opened 2024-03-14
Update sql-identifiers.md

### LINKED ISSUE #1930: Change required: Clarify naming conflict between unqualified columns and parentheses-less system functions
### Related code PR

None.

https://docs.risingwave.com/docs/current/sql-identifiers/

Under `Naming restrictions`, raise awareness of the following issue and provide workarounds:

In an **expression**, certain names are interpreted as builtins rather than column name. For example:

```
select user; -- This is the builtin `user`

select user, avatar from t; -- This is also the builtin `user`, rather than a column from the table `t`

select t.user, avatar from t; -- Qualify it with `t.` to select the column rather than the builtin.
```

The full list of such builtins include `user`, `current_timestamp`, `current_schema`, `current_role`, `current_user`, `session_user`. (Message to doc team: Full list [here](https://github.com/risingwavelabs/risingwave/blob/v1.7.0/src/frontend/src/binder/expr/function.rs#L48) but we may [extend](https://github.com/postgres/postgres/blob/REL_16_1/src/include/nodes/primnodes.h#L1451) it in future versions. The list can be hard to maintain and may not be necessary, as the most confusing one is just `user`.)

You can either avoid naming columns with these words, or qualify it with table name (as shown in the example) when such ambiguity happens.

(Feel free to rephrase / reorganize the content above.)

### Reference

https://github.com/risingwavelabs/risingwave/issues/15477

ShanlanLi:
## Info

- **Description**

  - [ What's changed? Which parts of the docs are affected? ]

- **Notes**

- **Related code PR**

- **Related doc issue**
  
  Resolves https://github.com/risingwavelabs/risingwave-docs/issues/1930

## For reviewers

- **Preview**

- **Key points**

## Before merging

### DIFF
--- a/docs/sql/sql-identifiers.md
+++ b/docs/sql/sql-identifiers.md
@@ -21,7 +21,7 @@ title: Identifiers
     SELECT user, avatar FROM t; -- This is also the builtin `user`, rather than a column from the table `t`.
     ```
 
-  Some of such names that need to be especially noticed include **`user`, `current_timestamp`, `current_schema`, `current_role`, `current_user`,** and **`session_user`**. To avoid such issues, you can either avoid naming a column with these words, or qualify it with the table name (as shown in the example below) when such ambiguity happens.
+  Several names also require special attention, including **`user`, `current_timestamp`, `current_schema`, `current_role`, `current_user`,** and **`session_user`**. To avoid such issues, you can either avoid naming a column with these words, or qualify it with the table name (as shown in the example below) when such ambiguity happens.
 
   ```sql title="Solution to avoid naming conflicts"
   SELECT t.user, avatar FROM t; -- Qualify it with `t.` to select the column rather than the builtin.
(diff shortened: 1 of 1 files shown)


xiangjinwu (approved):
Thank you!

docs/sql/sql-identifiers.md
```diff
@@ -13,7 +13,19 @@ title: Identifiers
...
 * You can circumvent any above rules by double-quoting the identifier (e.g., `"5_source"`). All characters inside a quoted identifier are taken literally, except double-quotes must be escaped by writing two adjacent double-quotes, as in (e.g., `"two""quotes"`).
+* In an **expression**, certain names are interpreted as builtins rather than column names. For example:
 
+    ```sql title="Names interpreted as builtins"
+    SELECT user; -- This is the builtin `user`.
+
+    SELECT user, avatar FROM t; -- This is also the builtin `user`, rather than a column from the table `t`.
+    ```
+
+  Some of such names that need to be especially noticed include **`user`, `current_timestamp`, `current_schema`, `current_role`, `current_user`,** and **`session_user`**. To avoid such issues, you can either avoid naming a column with these words, or qualify it with the table name (as shown in the example below) when such ambiguity happens.
```
WanYixian:
```suggestion
  Several names also require special attention, including **`user`, `current_timestamp`, `current_schema`, `current_role`, `current_user`,** and **`session_user`**. To avoid such issues, you can either avoid naming a column with these words, or qualify it with the table name (as shown in the example below) when such ambiguity happens.
```

WanYixian (approved):
Edit a little bit for clarity, others LGTM

PR #1994, opened 2024-03-21
Guide for ingesting CDC data from Neon

fahadullahshah261:
I created this PR regarding the guide for ingesting CDC data from Neon. Please, review it. Thanks.

PR #2017, opened 2024-03-28
Revert "Guide for ingesting CDC data from Neon"

hengm3467:
Reverts risingwavelabs/risingwave-docs#1994

We need to update the sidebar so that Neon shows up


hengm3467:
I'll update the sidebar in a separate PR.

PR #2167, opened 2024-05-12
remove faq about Kafka consumer group ID

xxchan:
Signed-off-by: xxchan <xxchan22f@gmail.com>

## Info

- **Description**

  - Remove the consumer group ID section, because it's misleading.
  - Update versioned doc: copy latest content to 1.8 (current). Show refer to latest for 1.7. Actually I think for FAQ, it does not need to be versioned, but I didn't find a good solution.

## For reviewers

- **Preview**

- **Key points**

## Before merging


docs/rw-faq.md
```diff
@@ -96,34 +96,6 @@ In contrast, materialized views in OLAP databases may not prioritize real-time u
...
-
-2. **Disruption of existing RisingWave behavior**: The issue arises when users are allowed to specify group IDs in sources. Let's consider a scenario where two downstream materialized views depend on the same source, assuming a parallelism degree of 3 and the upstream topic having 3 partitions. In the current implementation, there would be two sets of source executors, totaling six, all sharing the same group ID.
-
-According to Kafka's behavior, "each consumer in a group is assigned one or more partitions to read messages from." However, if the number of consumers exceeds the number of partitions, some consumers will not receive any data. This situation clearly fails to meet the requirement of both materialized views receiving complete data sets.
```
xxchan, line 125: This section is wrong. There will never be any data loss, since we are using Kafka `assign` API. The only difference is Kafka broker side monitoring.

references:
- https://github.com/risingwavelabs/risingwave/pull/16111
- [notion.so/risingwave-labs/Notes-on-Kafka-Consumer-Consumer-Group-6bafdbab58b34ad7917fe47645f9a862](https://www.notion.so/risingwave-labs/Notes-on-Kafka-Consumer-Consumer-Group-6bafdbab58b34ad7917fe47645f9a862)

xxchan: Other sections are correct, but I think they do not need to documented after this part is removed.

versioned_docs/version-1.7/rw-faq.md
```diff
@@ -8,32 +8,4 @@ slug: /rw-faq
...
-
-2. **Disruption of existing RisingWave behavior**: The issue arises when users are allowed to specify group IDs in sources. Let's consider a scenario where two downstream materialized views depend on the same source, assuming a parallelism degree of 3 and the upstream topic having 3 partitions. In the current implementation, there would be two sets of source executors, totaling six, all sharing the same group ID.
-
-According to Kafka's behavior, "each consumer in a group is assigned one or more partitions to read messages from." However, if the number of consumers exceeds the number of partitions, some consumers will not receive any data. This situation clearly fails to meet the requirement of both materialized views receiving complete data sets.
+<a href="https://docs.risingwave.com/docs/current/rw-faq/"> Please refer to the current version of the doc.</a>
```
ShanlanLi: 1.7 faq only has one question. If this one question is deleted, we may delete this file(then the sidebar setting needs to be changed too)? 

@WanYixian what do you think?

WanYixian: Yes we can delete 1.7 FAQ file and change the sidebar

xxchan: IMO ideally all versions should have the same content for the FAQ page (and also some others like introduction). How do you think?

WanYixian: Oh got your point, it actually doesn't need to be iterated, each version can be the same.

WanYixian (approved):
LGTM now.

tabVersion (approved):
LGTM on the motivation, we can now share the same group-id with one fragment but still not allow users to configure themselves.

KeXiangWang:
Can we still provide a reason why RW does not accept Kafka consumer group ID? Since it's a frequently asked question, if the old answer is inaccurate, we'd better correct it instead of eliminating it.

xxchan:
We can support it. Btw, where is it asked? I don’t find it’s frequently asked

KeXiangWang:
> Btw, where is it asked? I don’t find it’s frequently asked

Frequently asked in our slack channel.

PR #2275, opened 2024-06-17
Document optimization of backfilling

### LINKED ISSUE #2219: Document optimization of backfilling by aligning the "key" of upstream and downstream MV
https://www.notion.so/risingwave-labs/huntclub-barrier-pie-up-60dea768f5b142f783f532383f2a944b?pvs=4#5e32e243097140f0b8765d72bf9f9c8b

WanYixian:
## Info

- **Description**

  - Add the MV on MV backfilling optimization into doc

- **Related doc issue**
  
  Resolves https://github.com/risingwavelabs/risingwave-docs/issues/2219

## For reviewers

- **Preview**

- **Key points**

## Before merging


docs/performance/performance-best-practices.md
```diff
@@ -30,6 +30,8 @@ When building a materialized view (MV), it is possible to build more MVs on top
 
 Users may have concerns if decomposing a complex pipeline into multiple MVs introduces additional performance overhead. We remark that more decomposition does not lead to more computation overhead but only more storage space. Since RisingWave is typically deployed with cheap object store on the public cloud to store large amounts of data, we generally consider it a less crucial factor in most cases. We are planning to introduce a new feature that allows users to remove an intermediate MV if it's considered unnecessary and poses a significant storage space concern.
 
+Please be aware that when the MV on MV is backfilling, the data from the upstream MV may not have good locality. To optimize this, we recommend modifying the SQL query so that the order key of the upstream table aligns with the group key of the downstream MV. This alignment improves the performance of the backfilling process and ensures better efficiency.
```
lmatz: We need an example to show `order key` and `group key`, and how users can fine these two things in SQL

lmatz: I suppose `order key of the upstream table` is the `stream key`:
```
dev=> explain create table t;
                                                       QUERY PLAN                                                       
------------------------------------------------------------------------------------------------------------------------
 StreamMaterialize { columns: [_row_id(hidden)], stream_key: [_row_id], pk_columns: [_row_id], pk_conflict: Overwrite }
 └─StreamRowIdGen { row_id_index: 0 }
   └─StreamUnion { all: true }
     └─StreamExchange { dist: HashShard(_row_id) }
       └─StreamDml { columns: [_row_id] }
         └─StreamSource
(6 rows)
```
's 
```
stream_key: [_row_id]
```

but what's the group key of the downstream MV? cc: @st1page

st1page: order key of the upstream table is not the stream key. It is the primary key prefix of the table.
Actually the `group key` here is the primary key of the state using by the downstream MV, It could be the group key of Agg/TopN or join key of join or other cases.

lmatz: > Actually the group key here is the primary key of the state using by the downstream MV

The first stateful operator in the query of the downstream MV that reads from the upstream table?

lmatz:
I would like to add two examples of this optimization


The first one:

Suppose we have a table `t` by `create table t(v1 int, v2 varchar, v3 timestamp);`. It has a huge amount of historical data, which means building a new MV on top of it requires backfilling all this historical data, which implies it is a long process.

Let's say we want to build an MV as `create materialized view m as select v1, sum(v2) from t group by v1`.
The problem with backfilling is that when reading from t to build m, it is possible that the order of v1 is random, leading to bad cache locality and triggering remote I/O, which decreases the performance.

However, if we have `create table t(v1 int primary key, v2 varchar)` instead or defining an intermediate materialized view `create materialized view tmp as select * from t order by v1`.

The storage of RW will order rows by v1, which means that rows with same v1 will be read consecutively. Therefore, cache locality is strong, which leads to better performance.

The second one:

The reasoning also applies to other queries such as 
`create materialized view m as select lag(partition by v1 order by v3) from t`.

If we can define v3 as a primary key or build an intermediate MV that orders by v3, backfilling materialized view `m` would be much faster as `lag` can append new data instead of ordering an array of data according to v3.

Conclusion:
Please check the primary key of the base table or the base materialized view by `explain`ing them and see if any prefixes of the primary key can benefit the newly created MV.



cc: @st1page  do you think these two examples make sense?

st1page:
> I would like to add two examples of this optimization
> 
> The first one:
> 
> Suppose we have a table `t` by `create table t(v1 int, v2 varchar, v3 timestamp);`. It has a huge amount of historical data, which means building a new MV on top of it requires backfilling all this historical data, which implies it is a long process.
> 
> Let's say we want to build an MV as `create materialized view m as select v1, sum(v2) from t group by v1`. The problem with backfilling is that when reading from t to build m, it is possible that the order of v1 is random, leading to bad cache locality and triggering remote I/O, which decreases the performance.
> 
> However, if we have `create table t(v1 int primary key, v2 varchar)` instead or defining an intermediate materialized view `create materialized view tmp as select * from t order by v1`.
> 
> The storage of RW will order rows by v1, which means that rows with same v1 will be read consecutively. Therefore, cache locality is strong, which leads to better performance.
> 
> The second one:
> 
> The reasoning also applies to other queries such as `create materialized view m as select lag(partition by v1 order by v3) from t`.
> 
> If we can define v3 as a primary key or build an intermediate MV that orders by v3, backfilling materialized view `m` would be much faster as `lag` can append new data instead of ordering an array of data according to v3.
> 
> Conclusion: Please check the primary key of the base table or the base materialized view by `explain`ing them and see if any prefixes of the primary key can benefit the newly created MV.
> 
> cc: @st1page do you think these two examples make sense?

LGTM, But I'm not very sure about the performance improvement help in the second scenario. It might require some experiments. I think we can use the first example in the document. What do you think?

lmatz:
Make sense

cc: @WanYixian let's merge the first example only, could you help review and polish it?

WanYixian:
> Make sense
> 
> cc: @WanYixian let's merge the first example only, could you help review and polish it?

Noted, will do

lmatz (approved):

PR #2348, opened 2024-07-05
Add workaround for high join amplification 

### LINKED ISSUE #2330: Document workaround for high join amplifcation
Let’s say there are a few join keys that can trigger high join amplification,

currently, it will lead to the congestion of the whole job because the high join amplification of a single key produces a large number of results and they have to be processed by a single actor of that join operator. 

Therefore, simply scaling does not work.

We can split the original MV into multiple MVs by adding filtering conditions to make the data of the same key distributed to more actors, (if the data pattern and workload allow us to do so)

Let’s say the original join condition is
```
select * from
     orders
inner join
    product_description
where
orders.product_id = product_description.product_id
```

and an update from stream product_description with `product_id=1` can match 100K rows from t1 because `product_id = 1` is a hot sale product

We can try to split the MV into multiple MVs:
1. one MV with filtering conditions such as `orders.user_id % 7 = 0`, and then do the join
2. the second MV with filtering conditions such as `orders.user_id % 7 = 1`
...
...
N. the nth MV with filtering conditions `orders.user_id % 7 = 1`.

7 in total etc.

(If the `user_id` is not a number that can be applied with module operation, it needs to be processed by some hash function first. In the case when the hash function can return a negative value, we remark that in both PostgreSQL and RW, -1 % 7 = -1 instead of 6, therefore we need to include more MVs where `hash_function(orders.user_id) % 7 = -1/-2/-3/-4/-5/-6))

Finally, 
we union all the results from all 7 MVs. If user_id is uniformly distributed given `product_id = 1`, the join amplification in each actor is reduced by a factor of 7.

It all depends on whether we can find such a user_id column, i.e. high cardinality, and as uniformly distributed as possible.

WanYixian:
## Info

- **Description**

  - Add workaround for high join amplification 

- **Related doc issue**
  
  Resolves https://github.com/risingwavelabs/risingwave-docs/issues/2330

## For reviewers

- **Preview**

- **Key points**

## Before merging

### DIFF
--- a/docs/troubleshoot/troubleshoot-high-latency.md
+++ b/docs/troubleshoot/troubleshoot-high-latency.md
@@ -35,9 +35,13 @@ Once you've pinpointed the bottleneck fragment, consider the following actions t
 
 - Increase the parallelism by adding more nodes into the cluster, or check the SQL query to see if there is any room for optimization.
 
-## Example: High latency caused by high join amplification
+## Example
 
-Using low-cardinality columns as equal conditions in joins can result in high join amplification, leading to increased latency. 
+High latency can be caused by high join amplification.
+
+### Identify high join amplification
+
+Using low-cardinality columns as equal conditions in joins can result in high join amplification, leading to increased latency.
 
 :::info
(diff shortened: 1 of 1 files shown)


lmatz (approved):
Thanks! LGTM

PR #2397, opened 2024-07-22
Update performance best practice

### LINKED ISSUE #2381: Document tips for speeding up backfilling
Backfilling is the process when a query processes all the existing or historical data. 

Suppose we have 100GB of data in a topic of the upstream Kafka and new data keeps coming into the topic. Once we define a query that ingests data from this topic, it will first process all of the existing data. We call this stage backfilling. After that, it will continue processing the new data.

Since RW always tries to ingest and process data as quickly as possible, the stage of backfilling tends to saturate the cluster resources.

The performance of backfilling often becomes worse when the query becomes more complex, i.e. a large MV/sink query that has many CTE/subqueries. The definition of such a query is not necessarily the most efficient way of backfilling a query if the cluster is overloaded.

The reason is intuitive: running a query with 10 CTE/subqueries is similar to running 10 different queries at the same time.

There can be serious competition for resources:
1. memory. There is not enough memory to hold the state of all the stateful operators in the query, which leads to a high cache miss rate and high remote IOPS.
2. CPU. The block cache/meta cache will be very busy bringing in and evicting data. The data needs to be deserialized before they can be used by operators, which consumes more CPU.
3. remote I/O bandwidth. Due to (1), RW may hit object storage(e.g. S3/GCS)'s rate limit more frequently.

We remark that the competition for resources is not just implying that each of the 10 queries gets 1/10 of total resources, so each of them is expected to deliver 1/10 of the performance, but the fact that extra overhead and limitation lead to even worse performance than 1/10.

We suggest that if high CPU utilization or high cache miss rate is observed during backfilling of a large complex query, decomposing it into several independent queries could be helpful to the performance.

-------
It is a performance tip

WanYixian:
## Info

- **Description**

  - Add how to speed up backfilling into performance

- **Related doc issue**
  
  Resolves https://github.com/risingwavelabs/risingwave-docs/issues/2381

## For reviewers

- **Preview**

- **Key points**

## Before merging

### DIFF
--- a/docs/performance/performance-best-practices.md
+++ b/docs/performance/performance-best-practices.md
@@ -21,7 +21,7 @@ We determine how the index should be created by checking:
 
 We remark that the incremental maintenance of indexes in RisingWave is similar to the incremental maintenance of materialized views but with minimal computation. Therefore, it is cost-effective to create indexes in RisingWave. We encourage users to detect the patterns in batch queries and create indexes if the pattern occurs frequently or/and a batch query is slow.
 
-## When to speed up backfilling
+## How to speed up backfilling
 
 Backfilling refers to the operation where a query processes all existing or historical data.
 
--- a/docs/performance/performance-best-practices.md
+++ b/docs/performance/performance-best-practices.md
@@ -37,7 +37,7 @@ The reason is straightforward: running a query with 10 CTEs or subqueries is sim
 
 Note that this resource competition doesn't simply mean each of the 10 queries gets 1/10 of the total resources. Extra overhead and limitations can lead to even worse performance than 1/10 of the original.
 
-If you observe high CPU utilization or high cache miss rates during backfilling of a large, complex query, we suggest decomposing it into several independent queries. This approach could significantly improve performance.
+If you observe high CPU utilization or high cache miss rates during backfilling of a large, complex query, we suggest decomposing it into several independent queries and creating each MV one by one. This approach could significantly improve performance.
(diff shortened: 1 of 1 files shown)


lmatz (approved):
LGTM

docs/performance/performance-best-practices.md
```diff
@@ -21,6 +21,24 @@ We determine how the index should be created by checking:
...
+
+The reason is straightforward: running a query with 10 CTEs or subqueries is similar to running 10 different queries simultaneously. This can lead to serious resource competition:
+
+- Memory: There may not be enough memory to hold the state of all stateful operators in the query, resulting in high cache miss rates and high remote I/O operations per second (IOPS).
+- CPU: The block cache and meta cache become very busy bringing in and evicting data. Data must be deserialized before operators can use it, consuming more CPU.
+- Remote I/O bandwidth: Due to memory constraints, RisingWave may hit object storage (e.g., S3/GCS) rate limits more frequently.
+
+Note that this resource competition doesn't simply mean each of the 10 queries gets 1/10 of the total resources. Extra overhead and limitations can lead to even worse performance than 1/10 of the original.
+
+If you observe high CPU utilization or high cache miss rates during backfilling of a large, complex query, we suggest decomposing it into several independent queries. This approach could significantly improve performance.
```
lmatz: "we suggest decomposing it into several independent queries and creating each materialized view one by one."

PR #2412, opened 2024-07-29
Update subscription cursor

WanYixian:
## Info

- **Description**

  - Update fetch from cursor part for subscription 

## For reviewers

- **Preview**

- **Key points**

## Before merging

### DIFF
--- a/versioned_docs/version-1.10/transform/subscription.md
+++ b/versioned_docs/version-1.10/transform/subscription.md
@@ -110,9 +110,11 @@ FETCH n FROM cursor_name;
 
 #### Order of the fetched data
 
-- For rows with different `rw_timestamp`, values are returned in the order the events occurred.
+- For data with different `rw_timestamp`, values are returned in the order the events occurred.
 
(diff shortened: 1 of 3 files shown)


versioned_docs/version-1.10/transform/subscription.md
```diff
@@ -96,14 +100,20 @@ Note that each time `FETCH NEXT FROM cursor_name` is called, it will return one
...
 
 ```sql
 FETCH n FROM cursor_name;
 ```
 
+#### Order of the fetched data
+
+- For rows with different `rw_timestamp`, values are returned in the order the events occurred.
+
+- For rows with the same `rw_timestamp`, the order matches the event sequence if the primary keys are the same; If primary keys differ, the order may not reflect the exact event sequence.
```
xxhZs: A note may be required the PK is the same as the dependent mv or table's，

xxhZs (approved):
lgtm

versioned_docs/version-1.10/transform/subscription.md
WanYixian: Updated

PR #2023, opened 2024-03-28
Hide the single-table MySQL/PG cdc source and use the shared source instead.

fuyufjh:
## Info

- **Description**

Hide the single-table MySQL/PG cdc source and use the shared source instead.

Preparing for deprecating the single-table MySQL/PG cdc source.

- **Notes**

  - [ Include any supplementary context or references here. ]

- **Related code PR**

  - [ Provide a link to the relevant code PR here, if applicable. ]

- **Related doc issue**
  
  Resolves [ Provide a link to the relevant doc issue here, if applicable. ]

## For reviewers

- **Preview**

  - [ Paste the preview link to the updated page(s) here. Edit this item after the preview site is ready. To find the updated pages, scroll down to locate and open the Amplify preview link and select the **dev** version of the documentation. ]

- **Key points**

  - [ Parts that may need revision or extra consideration. ]

## Before merging

- [ ] I have checked the doc site preview, and the updated parts look good.

- [ ] I have acquired the approval from the owner (and optionally the reviewers) of the code PR and at least one tech writer (`CharlieSYH`, `emile-00`, & `hengm3467`).

PR #1754, opened 2024-01-11
Doc shared CDC source transactions and refactor example section

emile-00:
## Info

- **Description**

 - For both MySQL and PG CDC: 
    - Add note on transaction for shared CDC source
    - Refactor examples section to include create single CDC table and create multiple tables from shared source

- **Notes**

  - [ Include any supplementary context or references here. ]

- **Related code PR**

  - https://github.com/risingwavelabs/risingwave/pull/14375

- **Related doc issue**
  
  Resolves https://github.com/risingwavelabs/risingwave-docs/issues/1732, https://github.com/risingwavelabs/risingwave-docs/issues/1726

## For reviewers

- **Preview**

  - [ Paste the preview link to the updated page(s) here. Edit this item after the preview site is ready. To find the updated pages, scroll down to locate and open the Amplify preview link and select the **dev** version of the documentation. ]

- **Key points**

  - [ Parts that may need revision or extra consideration. ]

## Before merging

- [ ] I have checked the doc site preview, and the updated parts look good.

- [ ] I have acquired the approval from the owner (and optionally the reviewers) of the code PR and at least one tech writer (`CharlieSYH`, `emile-00`, & `hengm3467`).

PR #2449, opened 2024-08-12
fix(source): PostgreSQL/MySQL CDC source vs table relationship

xiangjinwu:
## Info

- **Description**

  - Initially, CDC can only be created as `table` but not `source`.
  - After [#1754](https://github.com/risingwavelabs/risingwave-docs/pull/1754), there are 2 options:
    - Create a `table` directly, same as before.
    - Create a shared `source` first, and then `table` based on the shared source.
  - After [#2023](https://github.com/risingwavelabs/risingwave-docs/pull/2023), the first syntax is deprecated and the second syntax is recommended.

- Fixes in this PR (using PostgreSQL as example but may apply to MySQL as well):

  - `table or source` is never the case. It is `table`(deprecated) or `source + table` (recommended).
  - Link to `create table` syntax reference is useless. That page does not contain `FROM source_name TABLE` syntax for shared PostgreSQL/MySQL CDC.
  - `FROM source TABLE` is updated to `FROM source_name TABLE` in the example, to keep consistent with `CREATE SOURCE source_name` above.

- **Notes**

  - Once @StrikeW confirmed the content above is accurate, we may apply the same fix from PostgreSQL to MySQL, and from dev version to current or even older applicable versions.

- **Related code PR**

- **Related doc issue**
  
## For reviewers

- **Preview**

- **Key points**

## Before merging


xiangjinwu:
Btw is the `|table.name| Name of the table that you want to ingest data from. |` parameter still useful? If the source is tied to an upstream table, there would be no point to **share** the source for multiple RisingWave tables, right?

StrikeW (approved):
Thanks for the fix.

docs/guides/ingest-from-mysql-cdc.md
```diff
@@ -149,11 +149,11 @@ If you are running RisingWave locally from binaries and intend to use the native
 
 ## Create a table using the native CDC connector in RisingWave
 
-To ensure all data changes are captured, you must create a table and specify primary keys. See the [`CREATE TABLE`](/sql/commands/sql-create-table.md) command for more details.
+To ensure all data changes are captured, you must create a table and specify primary keys.
 
 ### Syntax
 
-Syntax for creating a CDC source.
+Syntax for creating a shared CDC source.
```
fuyufjh: Better to clarify the logic first: need to create a `source` before creating `table`

```suggestion
Before creating a table, you need to first create a source to the upstream MySQL database using the following syntax. This source can be shared by multiple tables from the same MySQL database.
```

fuyufjh (approved):
Overall LGTM

docs/guides/ingest-from-mysql-cdc.md
```diff
@@ -162,7 +162,7 @@ CREATE SOURCE [ IF NOT EXISTS ] source_name WITH (
 );
 ```
 
-Syntax for creating a CDC table. Note that a primary key is required and must be consistent with the upstream table.
+Syntax for creating a CDC table based on the shared source. Note that a primary key is required and must be consistent with the upstream table.
```
hengm3467: Better use a complete sentence. 
```suggestion
Below is the syntax for creating a CDC table based on the shared source. Note that a primary key is required and must be consistent with the upstream table.
```

docs/guides/ingest-from-postgres-cdc.md
```diff
@@ -151,11 +151,11 @@ If you are running RisingWave locally from binaries and intend to use the native
 
 ## Create a table using the native CDC connector
 
-To ensure all data changes are captured, you must create a table or source and specify primary keys. See the [`CREATE TABLE`](/sql/commands/sql-create-table.md) command for more details.
+To ensure all data changes are captured, you must create a shared source first and create tables based on that source.
```
hengm3467, line 162: It's ok as of now in this PR, but I feel it would be more useful if we have a topic or section about "shared sources" that we can now point to. There are a few RisingWave-specific concepts that we may want to elaborate somewhere. In addition, the "Sources" term in Glossary should be updated. cc: @st1page @WanYixian

st1page: Good catch. The shared source is still in development by @xxchan  and it is expected to release in v2.1. I think we will add the doc that time.

hengm3467: I see. Then let's merge this PR and handle that separately.

PR #2474, opened 2024-08-16
link to latest stable version from unmaintained and unreleased version

stdrc:
## Description

Previously we have a customized `DocVersionBanner` which only links to latest from "next", while actually Docusaurus natively supports this linking feature. This PR removes the custom `DocVersionBanner` and adjusted the `versions` config to make it work.

## Checklist


docusaurus.config.js
```diff
@@ -32,78 +32,67 @@ const config = {
           sidebarCollapsible: false,
           showLastUpdateAuthor: false,
           showLastUpdateTime: true,
+          lastVersion: "1.10",
           versions: {
             current: {
-              label: "1.11 (dev)",
-              path: "/dev",
+              label: "next",
```
stdrc, line 38: We don't need to mention the version number of unreleased version IMO, to prevent versioning policy change like `1.11 -> 2.0` this time.

WanYixian: Makes sense!

WanYixian (approved):

PR #2481, opened 2024-08-19
Made udaf more visible in overview page

### LINKED ISSUE #2476: Change required: UDAF should be mentioned in UDF docs
### Related code PR

There is a [dedicated series of documents](https://docs.risingwave.com/docs/current/user-defined-functions/) for UDF, which is good. 

image

However, the UDAF is never mentioned in these docs. Conceptually, UDAF is very close to UDF, or you can consider it as a subset of UDF. I think we need to merge them together.

Note that UDAF is mentioned here: https://docs.risingwave.com/docs/current/sql-create-aggregate now. Even though, according to https://github.com/risingwavelabs/risingwave-docs/issues/2442, both the reference **and** the guide should exist.

- https://docs.risingwave.com/docs/current/user-defined-functions Needs to be enhanced with UDAF contents
- https://docs.risingwave.com/docs/current/sql-create-aggregate Needs to hyperlink to that page

### Reference

_No response_

ShanlanLi:
## Description

As title.

## Related doc issue

Resolves https://github.com/risingwavelabs/risingwave-docs/issues/2476

## Rendered preview

https://pr-2481.d2fbku9n2b6wde.amplifyapp.com/docs/next/user-defined-functions/

## Checklist


fuyufjh (approved):
LGTM!

But for #2476, I think more changes need to be done for 

- https://docs.risingwave.com/docs/current/udf-python/
- https://docs.risingwave.com/docs/current/udf-java/
- https://docs.risingwave.com/docs/current/udf-python-embedded/
- https://docs.risingwave.com/docs/current/udf-javascript/

Particularly, adding instructions and code examples about how to write UDAF.

WanYixian (approved):
LGTM

PR #2505, opened 2024-08-23
move faq as top level section

st1page:
## Description

As title. 
Before the `FAQ` was placed in `About`, it took on a large part of helping users understand the usefulness of risingwave. But at the same time after that we added a lot of questions from users in real scenarios. This PR moves the FAQ out to focus on the latter requirement. This is also due to the fact that some of the content in the current FAQ is outdated due to a change in the promotion and positioning of our product. We may also add new pages in the future to feature risingwave.

## Checklist

### DIFF
--- a/docs/faq/overview.md
+++ b/docs/faq/overview.md
@@ -9,6 +9,6 @@ slug: /faq-overview
 
 In this section, we present the most frequently asked questions and some additional information.
 
-### Have questions that weren’t answered?
+### Questions not covered in the FAQ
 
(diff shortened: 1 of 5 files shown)


docs/faq/overview.md
```diff
@@ -0,0 +1,14 @@
...
+title: Overview
+slug: /faq-overview
+---
+<head>
+  <link rel="canonical" href="https://docs.risingwave.com/docs/current/faq-using-risingwave/" />
+</head>
+
+In this section, we present the most frequently asked questions and some additional information.
+
+### Have questions that weren’t answered?
```
hengm3467:
```suggestion
### Questions not covered in the FAQ
```

hengm3467 (approved):
Otherwise LGTM. Thanks!

PR #2519, opened 2024-08-27
2363 document how to tune reserved memory and cache eviction policy

### LINKED ISSUE #2363: Document how to tune reserved memory and cache eviction policy
This is an advanced material and we don't expect most users to adjust it. 
We can tune it for RW Cloud users as we have access to the Grafana dashboards. 
But anyway, we can offer some guidelines for open-sourced users.

https://docs.risingwave.com/docs/current/performance-metrics/#memory-usage
One thing that needs to be fixed is that the portion of reserved memory is changed in the kernel. This needs to be aligned with: https://github.com/risingwavelabs/risingwave-docs/pull/2312. I suggest adding a link to it.

https://docs.risingwave.com/docs/dev/rw-faq/#why-is-the-memory-for-compute-nodes-not-fully-utilized
The paragraph told the users that they can adjust reserved memory by `RW_RESERVED_MEMORY_BYTES=8589934592 or --reserved-memory-bytes=8589934592`.

Another factor that affects how much memory users would use is the following configuration variables set in the configuration `toml` file:
```
    #[serde(default = "default::developer::memory_controller_threshold_aggressive")]
    pub memory_controller_threshold_aggressive: f64,

    #[serde(default = "default::developer::memory_controller_threshold_graceful")]
    pub memory_controller_threshold_graceful: f64,

    #[serde(default = "default::developer::memory_controller_threshold_stable")]
    pub memory_controller_threshold_stable: f64,

    #[serde(default = "default::developer::memory_controller_eviction_factor_aggressive")]
    pub memory_controller_eviction_factor_aggressive: f64,

    #[serde(default = "default::developer::memory_controller_eviction_factor_graceful")]
    pub memory_controller_eviction_factor_graceful: f64,

    #[serde(default = "default::developer::memory_controller_eviction_factor_stable")]
    pub memory_controller_eviction_factor_stable: f64,
```

In essence, `memory_controller_eviction_factor_XXX` decides:
1. if the memory usage exceeds `memory_controller_threshold_stable` of the total memory, the memory management mechanism in RW will start to evict data in the operator cache. But the intensity of eviction is mild at the moment.  
2. If the memory usage exceeds `memory_controller_threshold_graceful` of the total memory, the intensity of eviction becomes more aggressive.
3. If the memory usage eventually exceeds `memory_controller_threshold_aggressive`, the intensity of eviction becomes most aggressive.

In terms of the details about how the policy adjusts the intensity, we recommend users to take a look at the code directly: https://github.com/risingwavelabs/risingwave/blob/main/src/compute/src/memory/controller.rs
Increasing the `eviction_factor` will increase the intensity of eviction.

Suppose the reserved memory is 30%, and the `memory_controller_threshold_stable` is 0.8. We expect RW can use 1.0*(1-30%)*0.8 = 0.56 = 56% stably during most of the time. Typically, only when there is a sudden throughput spike or when creating a new materialized view that backfills a large amount of historical data, the memory usage will go beyond 56% and trigger more aggressive memory eviction. 

The best configuration highly depends on the workload and the data pattern. We recommend users tune the parameters based on the metrics on Grafana.

ShanlanLi:
## Related doc issue

Resolves https://github.com/risingwavelabs/risingwave-docs/issues/2363

## Rendered preview
https://pr-2519.d2fbku9n2b6wde.amplifyapp.com/docs/next/tune-reserved-memory/

## Checklist

### DIFF
--- a/docs/deploy/tune-reserved-memory.md
+++ b/docs/deploy/tune-reserved-memory.md
@@ -59,7 +59,7 @@ For instance, suppose you are deploying a compute node on a machine or pod with 
 
 - The total reserved memory would be 4.8GB + 9.6GB, which equals 14.4GB.
 
-However, if you find this excessive for your specific use case, you have the option to specify a different value. You can set either `RW_RESERVED_MEMORY_BYTES=8589934592` or `--reserved-memory-bytes=8589934592` when starting up the compute node. This will allocate 8GB as the reserved memory instead.
+However, if you find this excessive for your specific use case, you have the option to specify a different value. You can set either `RW_RESERVED_MEMORY_BYTES=8589934592` as an environment variable or `--reserved-memory-bytes=8589934592` as a command line parameter when starting up the compute node. This will allocate 8GB as the reserved memory instead.
 
 <details>
 <summary>Reserved memory setting before 1.10</summary>
--- a/docs/deploy/tune-reserved-memory.md
+++ b/docs/deploy/tune-reserved-memory.md
@@ -1,6 +1,6 @@
 ---
 id: tune-reserved-memory
-title: Tune reserved memory
+title: Tune reserved memory and cache eviction policy
 description: Tune the reserved memory and cache eviction policy in RisingWave.
 slug: /tune-reserved-memory
 ---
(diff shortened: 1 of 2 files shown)


docs/deploy/tune-reserved-memory.md
```diff
@@ -0,0 +1,126 @@
...
+
+For instance, suppose you are deploying a compute node on a machine or pod with 64GB of memory. By default, the reserved memory would be calculated as follows:
+
+- 30% of the first 16GB is 4.8GB.
+
+- 20% of the remaining 48GB (64GB - 16GB) is 9.6GB.
+
+- The total reserved memory would be 4.8GB + 9.6GB, which equals 14.4GB.
+
+However, if you find this excessive for your specific use case, you have the option to specify a different value. You can set either `RW_RESERVED_MEMORY_BYTES=8589934592` or `--reserved-memory-bytes=8589934592` when starting up the compute node. This will allocate 8GB as the reserved memory instead.
```
lmatz:
```suggestion
However, if you find this excessive for your specific use case, you have the option to specify a different value. You can set either `RW_RESERVED_MEMORY_BYTES=8589934592` as an environment variable or `--reserved-memory-bytes=8589934592` as a command line parameter when starting up the compute node. This will allocate 8GB as the reserved memory instead.
```

lmatz (approved):

docs/deploy/tune-reserved-memory.md
```diff
@@ -0,0 +1,126 @@
...
+---
+<head>
+  <link rel="canonical" href="https://docs.risingwave.com/docs/current/tune-reserved-memory/" />
+</head>
+
+This topic introduces the concept and the tuning method of reserved memory and cache eviction policy.
+
+## What is reserved memory and cache eviction policy?
+
+A simplified description of our memory control mechanism will help you understand these two concepts more straightforward. In simple words, RisingWave manages memory usage through a three-step process:
```
WanYixian:
```suggestion
A simplified description of our memory control mechanism will help you understand these two concepts more straightforwardly. In simple words, RisingWave manages memory usage through a three-step process:
```

docs/deploy/tune-reserved-memory.md
```diff
@@ -0,0 +1,126 @@
...
+
+1. Monitoring: The system calculates and monitors the memory usage of each component.
+
+2. Reserving buffer: The system reserves 30% of the total memory as a buffer. This is the **reserved memory**. In case of a sudden spike in input data, RisingWave has enough time to adjust its memory usage.
+
+3. Eviction checks: The system regularly compares current memory use against the remaining 70% (usable memory), and decides if it should evict data. This  **cache eviction policy** is based on usage levels:
```
WanYixian:
```suggestion
3. Eviction checks: The system regularly compares current memory usage against the remaining 70% (usable memory), and decides if it should evict data. RisingWave applies a **cache eviction policy** based on the memory usage levels:
```

WanYixian (approved):
Others LGTM!

PR #2547, opened 2024-08-30
refactor(theme): move custom React components to `src/components`

stdrc:
## Description

Changes:

1. Move custom React components to `src/components`
2. Manually import components when needing them in MDX files
3. Change `DefaultButton` and `LightButton` to use module-scoped CSS
4. Fix admonitions broken in last PR.

## Checklist


Ad-Bean (approved):
LGTM, I think the importing rules are changed in docusaurus 3.0+

WanYixian:
Hi Richard, I just found the admonition label is invalid, maybe because these are removed previously, shall we add them back? Otherwise, all the tip, note, warning would be just a paragraph without background color. Thanks!
image
image

stdrc:
> Hi Richard, I just found the admonition label is invalid, maybe because these are removed previously, shall we add them back? Otherwise, all the tip, note, warning would be just a paragraph without background color. Thanks!
> image
> image

Thanks for pointing this out. Let me fix it.

WanYixian (approved):

PR #2549, opened 2024-08-30
fix(style): fix image height

stdrc:
## Description

This PR fixes a image height bug introduced in the last PR.

Before:

截屏2024-08-30 22 16 49

After:

截屏2024-08-30 22 16 53

## Checklist


WanYixian (approved):
Just noticed another image has the same issue, so the update in CSS would fix this issue once and for all, right
image

stdrc:
> Just noticed another image has the same issue, so the update in CSS would fix this issue once and for all, right

Yes. This was my mistake. This PR fixes them all.

PR #2553, opened 2024-08-31
fix(md): fix ill-formed links in markdown files

stdrc:
## Description

Some broken links in our docs are blocking upgrading to Docusaurus v3. This PR fixes them.

Markdown guide from this PR:

- Use relative paths of `.md` files when doing cross-reference if possible, e.g. `[Explore RisingWave with examples](./quickstart.md#step-4-explore-risingwave-with-examples)`.
- No need to add `/` after the `.md` file path. ❌ `./get-started.md/#run-risingwave` ✅ `./get-started.md#run-risingwave` (the latter one can be clicked-into in VS Code)
- No need to maintain links to doc pages in Release Notes, because cross-referenceing by relative path is not supported between `pages` and `docs` directories. If using `/docs/current/...`, we must update these links each time we release a new version, and that's a burden.

## Checklist

### DIFF
--- a/cloud/console-overview.md
+++ b/cloud/console-overview.md
@@ -69,7 +69,7 @@ width="270px"
 
 The sample queries cover the most common steps in RisingWave, such as establishing a connection with a data source, processing data by defining materialized views and querying the results.
 
-See [Explore RisingWave with examples](/quickstart.md/?step=4) for details.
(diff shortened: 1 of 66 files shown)


WanYixian (approved):
Many thanks

PR #2563, opened 2024-09-02
Revamp transform data

st1page:
## Description
- rename "transform & query data" to "processing data"
- modify the overview of the processing data
- refactor the catagories in this part
- move "subscription" in "deliever data"

Current sidebar index 
image

## Related doc issue

Resolve

## Checklist


docs/transform/overview.md
```diff
@@ -7,13 +7,39 @@ title: Overview
...
+**Streaming**: RisingWave allows users to predefine SQL queries with `CREATE MATERIALIZED VIEW` statement. Risingwave continuously listens changes in upstream tables(in the `FROM` clause) and incrementally update the results automatically.
+
+**Batch**: Also like traditional databases, RisingWave allows users to send `SELECT` statement to query the result. At this point, RisingWave reads the data from the current snapshot, processes it, and returns the results.
+
+![Stream processing v.s. batch processing](../images/stream_processing_vs_batch_processing.png)
+
+
+Both modes have their unique advantages. Here are some considerations:
+
+**Cost & Performance**: Streaming mode can pre-compute and store results, which may improve query performance since the heavy lifting is done upfront.
```
hzxa21: Will this leave an impression that streaming is expensive to our user? I am thinking whether it is better to replace `Cost & Performance` with `Freshness`.

st1page: What is actually being considered here is that the streaming approach actually eliminates the need for duplicate computation over each ad hoc batch query, and therefore has a lower cost.

hzxa21: I see. I mistakenly thought we are mentioning `Cost v.s. Performance` as a trade-off here instead of `Cost & Performance`. If having a lower cost is the advantage, I suggest we mention it explicitly like:
```
Streaming mode can pre-compute and store results, which may improve query performance with lower cost since the heavy lifting is done upfront.
```

docs/transform/overview.md
```diff
@@ -7,13 +7,39 @@ title: Overview
...
+**Powerful**: RisingWave fully supports and optimizes a variety of SQL features, including advanced features such OVER window and different kinds of JOINs. At the same time, we are also committed to expanding the expressive power of SQL, such as by adding semi-structured data types and corresponding expressions.
+
+## When is Data Processing Performed? Batch(On Read) v.s. Streaming(On Write)
+
+There are 2 execution modes in our system serving different analytics purposes. The results of these two modes are the same and the difference lies in the timing of data processing, whether it occurs at the time of data ingestion(on write) or when the query is executed(on read).
+
+**Streaming**: RisingWave allows users to predefine SQL queries with `CREATE MATERIALIZED VIEW` statement. Risingwave continuously listens changes in upstream tables(in the `FROM` clause) and incrementally update the results automatically.
+
+**Batch**: Also like traditional databases, RisingWave allows users to send `SELECT` statement to query the result. At this point, RisingWave reads the data from the current snapshot, processes it, and returns the results.
```
fuyufjh: Shall we use the word **"ad-hoc query"** instead of "batch" across the documents to keep terminology consistent? Data warehouses usually don't use such word, and users may be confused with it, unless he has a good background on Big Data systems.

Only we (RisingWave developers) are using "batch processing" because it's often mentioned as a contract to "streaming processing".

docs/transform/overview.md
```diff
@@ -7,13 +7,39 @@ title: Overview
...
+
+![Stream processing v.s. batch processing](../images/stream_processing_vs_batch_processing.png)
+
+
+Both modes have their unique advantages. Here are some considerations:
+
+**Cost & Performance**: Compared to traditional databases, streaming mode can pre-compute and store results. The heavy lifting is done upfront, eliminating duplicate computation over each ad hoc query, and therefore has better performance and a lower cost.
+
+**Flexibility**: The streaming mode is less flexible to changes in query requirements. Especially for ad-hoc queries, the batch mode is still necessary.
```
fuyufjh: Ditto. Keep in mind that "Batch mode" is a word that ourselves invented.

sidebars.js
fuyufjh, line 1: So here is the new structure

```
Process data
    Overview
    Continuous streaming pipeline
        Dynamic filters
        Temporal filters
        Maintain wide table with table sinks
        Watermarks
        Emit on window close
    Ad-hoc batch query
        Indexes
        Time travel queries
    SQL features
        Joins
        Time windows
        Top-N by group
        Deletes and updates
```

As a user, at the first glance, I can't get the point that "SQL features" is the **common part shared** between "Continuous streaming pipeline" and "Ad-hoc batch query".

This is a critical fault and makes me -1 for the new arrangement. :( 

-----

I am thinking that perhaps the labeling approach from Flink is better than this. For [example](https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/dev/table/sql/queries/joins/)

image

which clearly shows a feature is for both or streaming-only or batch-only.

sidebars.js
```diff
@@ -351,77 +351,85 @@ const sidebars = {
...
+                  id: "transform/indexes",
+                  label: "Indexes",
+                },
+                {
+                  type: "doc",
+                  id: "transform/time-travel-queries",
+                  label: "Time travel queries",
+                },
+
+              ]
```
chenzl25: Can we add a link [Query using SELECT](https://docs.risingwave.com/docs/current/sql-select/) to this section?

st1page: I've just removed the previous categorization by batch and streaming. I'd like to add the link in the overview

st1page:
Engineer's review has been done PTAL cc @ShanlanLi @WanYixian

WanYixian (approved):
LGTM

ShanlanLi (approved):
My review done, this new transform overview looks good, thanks

PR #2582, opened 2024-09-06
refactor(style): refine sidebar styles

### LINKED ISSUE #2501: Enhancement: style of the sidebar navigation bar
Currently, in terms of styling, I think we can improve from two aspects:

1. The hierarchy is not clear, especially when there is deep nesting. I believe we can enhance clarity by adding some indentation or vertical lines.
2. The information density is relatively low. Compared to other doc site, there is excessive spacing between items, particularly around the first-level headings. Additionally, I think we can widen the sidebar to reduce line breaks.

image
image

stdrc:
## Description

1. Adjust indent width and hover bg color width for sidebar items.
2. Increase sidebar width.
3. Decrease vertical paddings of sidebar items.
4. Fix light/dark switch button hover color.
5. Change light theme accent color to use the color of RW logo.

Before

截屏2024-09-06 15 35 38

After:

截屏2024-09-06 15 35 12

## Related doc issue

Resolve #2501

## Checklist


WanYixian (approved):
LGTM!
BTW, I found the "ASK AI" function on the right side is blocking our table of contents a bit. Could we move it to a different position? Moving it further up or down would also obstruct other info element. 
image
What about moving it together with search to the middle since their functions are to some extent similar, e.g.:
image

st1page (approved):
LGTM

st1page:
> TW, I found the "ASK AI" function on the right side is blocking our table of contents a bit. Could we move it to a different position? Moving it further up or down would also obstruct other info element.

Some reference
image

image

PR #2637, opened 2024-09-25
Update generate_series example to use `NOW()::timestamp` since timestamptz is not supported for generate_series

kwannoel:
## Description

generate_series only works with timestamp, not timestamptz. The example in our doc fails due to this. Perhaps some regression? Not sure if `NOW()` changed to return `timestamptz` instead of `timestamp`.

```
dev=> create materialized view m1 as SELECT time::timestamptz FROM generate_series(
  '2020-01-01 00:00:00',
  now(),
  interval '1 hour'
) t(time);
ERROR:  Failed to run the query

Caused by:
  function generate_series(unknown, timestamp with time zone, interval) does not exist


dev=> create materialized view m1 as SELECT * FROM generate_series(
  '2020-01-01 00:00:00'::timestamptz,
  now(),
  interval '1 hour'
);
NOTICE:  Your session timezone is UTC. It was used in the interpretation of timestamps and dates in your query. If this is unintended, change your timezone to match that of your data's with `set timezone = [timezone]` or rewrite your query with an explicit timezone conversion, e.g. with `AT TIME ZONE`.

CREATE_MATERIALIZED_VIEW
```

## Related doc issue

Resolve

## Checklist

### DIFF
--- a/docs/sql/functions-operators/sql-function-set-returning.md
+++ b/docs/sql/functions-operators/sql-function-set-returning.md
@@ -89,7 +89,7 @@ Except for generating a static set of values, RisingWave also supports continuou
 CREATE MATERIALIZED VIEW mv AS
 SELECT * FROM generate_series(
   '2020-01-01 00:00:00',
-  now(),
+  now()::timestamp,
   interval '1 hour'
 );
 ```
--- a/docs/sql/functions-operators/sql-function-set-returning.md
+++ b/docs/sql/functions-operators/sql-function-set-returning.md
@@ -88,8 +88,8 @@ Except for generating a static set of values, RisingWave also supports continuou
 ```sql
 CREATE MATERIALIZED VIEW mv AS
 SELECT * FROM generate_series(
-  '2020-01-01 00:00:00',
-  now()::timestamp,
+  '2020-01-01 00:00:00'::timestamptz,
+  now(),
   interval '1 hour'
 );
 ```
--- a/versioned_docs/version-2.0/sql/functions-operators/sql-function-set-returning.md
+++ b/versioned_docs/version-2.0/sql/functions-operators/sql-function-set-returning.md
@@ -88,7 +88,7 @@ Except for generating a static set of values, RisingWave also supports continuou
 CREATE MATERIALIZED VIEW mv AS
 SELECT * FROM generate_series(
   '2020-01-01 00:00:00',
-  now(),
+  now()::timestamp,
   interval '1 hour'
 );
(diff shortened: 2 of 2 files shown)


stdrc:
Streaming `generate_series` with `now()` as end doesn't go through the normal function binding path. The example in doc is streaming query, not batch query, though this is indeed a bit confusing.

kwannoel:
> Streaming `generate_series` with `now()` as end doesn't go through the normal function binding path. The example in doc is streaming query, not batch query, though this is indeed a bit confusing.


```
dev=> create materialized view m1 as SELECT time::timestamptz FROM generate_series(
  '2020-01-01 00:00:00',
  now(),
  interval '1 hour'
) t(time);
ERROR:  Failed to run the query

Caused by:
  function generate_series(unknown, timestamp with time zone, interval) does not exist

```

Is it a regression then? 🤔

stdrc:
> > Streaming `generate_series` with `now()` as end doesn't go through the normal function binding path. The example in doc is streaming query, not batch query, though this is indeed a bit confusing.
> 
> ```
> dev=> create materialized view m1 as SELECT time::timestamptz FROM generate_series(
>   '2020-01-01 00:00:00',
>   now(),
>   interval '1 hour'
> ) t(time);
> ERROR:  Failed to run the query
> 
> Caused by:
>   function generate_series(unknown, timestamp with time zone, interval) does not exist
> ```
> 
> Is it a regression then? 🤔

Interesting... Let me fix this tmr🤣

kwannoel:
> > > Streaming `generate_series` with `now()` as end doesn't go through the normal function binding path. The example in doc is streaming query, not batch query, though this is indeed a bit confusing.
> > 
> > 
> > ```
> > dev=> create materialized view m1 as SELECT time::timestamptz FROM generate_series(
> >   '2020-01-01 00:00:00',
> >   now(),
> >   interval '1 hour'
> > ) t(time);
> > ERROR:  Failed to run the query
> > 
> > Caused by:
> >   function generate_series(unknown, timestamp with time zone, interval) does not exist
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is it a regression then? 🤔
> 
> Interesting... Let me fix this tmr🤣

Okay no rush, you can take a look first. I updated the doc accoridngly. Thanks for clarifying it's in streaming, I missed that part.

stdrc (approved):

WanYixian (approved):

PR #2660, opened 2024-09-30
fix: remove schema.registry.name.strategy

tabVersion:
## Description

we do not need `schema.registry.name.strategy` because now we no longer get key schema from schema registry.
Keeping the default behavior to `<topic name>-value` as value schema subject 

## Related doc issue

Resolve

## Checklist


docs/ingest/supported-sources-and-formats.md
```diff
@@ -64,7 +58,6 @@ Syntax:
 FORMAT PLAIN
 ENCODE AVRO (
    schema.registry = 'schema_registry_url [, ...]',
-   [schema.registry.name.strategy = 'topic_name_strategy'],
    [message = 'main_message'],
```
xiangjinwu: For avro (but not protobuf), `message` is only useful when `record_name_strategy` or `topic_record_name_strategy`. It shall be removed as well.

docs/ingest/supported-sources-and-formats.md
```diff
@@ -96,7 +87,6 @@ FORMAT DEBEZIUM
 ENCODE AVRO (
    message = 'main_message',
    schema.location = 'location' | schema.registry = 'schema_registry_url [, ...]',
-   [schema.registry.name.strategy = 'topic_name_strategy'],
    [key.message = 'test_key'],
```
xiangjinwu: Same here. Let's remove both `message` and `key.message` for Debezium Avro.

Btw `schema.location` shall not be here for avro either. It has to be `schema.registry`.

docs/ingest/supported-sources-and-formats.md
```diff
@@ -108,15 +98,12 @@ When consuming data in AVRO from Kafka topics, the `FORMAT` and `ENCODE` section
...
-Optionally, you can define a `schema.registry.name.strategy` if `schema.registry` is set. Accepted options include `topic_name_strategy`, `record_name_strategy`, and `topic_record_name_strategy`. If either `record_name_strategy` or `topic_record_name_strategy` is used, the `message` field must also be defined. For additional details on name strategy, see [Subject name strategy](https://docs.confluent.io/platform/current/schema-registry/fundamentals/serdes-develop/index.html#subject-name-strategy).
-
 Syntax:
 
 ```sql
 FORMAT UPSERT
 ENCODE AVRO (
    schema.location = 'location' | schema.registry = 'schema_registry_url [, ...]',
-   [schema.registry.name.strategy = 'topic_name_strategy'],
    [message = 'main_message'],
```
xiangjinwu: `message` no longer useful

xiangjinwu (approved):

WanYixian (approved):
Since which version does this change apply to, 2.0 or 2.1? 🤔

tabVersion:
> Since which version does this change apply to, 2.0 or 2.1? 🤔

We can remove it since 2.0.

WanYixian (approved):
LGTM

PR #2676, opened 2024-10-12
Fix input type error

devprofile98:
## Description
in `CREATE FUNCTION key_value(bytea) RETURN ...` , the function input variable type is wrong, it should be '`varchar`' instead of `bytea` as declared here:
``` python
@udf(input_types=['VARCHAR'], result_type='STRUCT<key: VARCHAR, value: VARCHAR>')
def key_value(pair: str):
```

## Related doc issue

Resolve

## Checklist


devprofile98:
@hengm3467 Hi, can you review this pr?

hengm3467 (approved):
LGTM. Thanks!

hengm3467:
Just noticed that the change was made to v2.0 docs. @WanYixian Could you please replicate the changes to the `next` version? Thanks.

PR #2713, opened 2024-10-22
Doc `SET TIME ZONE INTERVAL '+00:00' HOUR TO MINUTE`

### LINKED ISSUE #2712: Document: feat: support SET TIME ZONE INTERVAL '+00:00' HOUR TO MINUTE
### Related code PR

https://github.com/risingwavelabs/risingwave/pull/18705

https://docs.risingwave.com/docs/next/sql-set-time-zone/

### Reference

_No response_

WanYixian:
## Description

Doc `SET TIME ZONE INTERVAL '+00:00' HOUR TO MINUTE` since v2.1

## Related code PR

https://github.com/risingwavelabs/risingwave/pull/18705

## Related doc issue

Resolve https://github.com/risingwavelabs/risingwave-docs/issues/2712

## Checklist

### DIFF
--- a/docs/sql/commands/sql-set-time-zone.md
+++ b/docs/sql/commands/sql-set-time-zone.md
@@ -23,7 +23,7 @@ SET TIME ZONE { time_zone | LOCAL | DEFAULT };
 
 | Parameter                 | Description           |
 | ------------------------- | --------------------- |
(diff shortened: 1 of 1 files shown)


chenzl25:
I think we don't need to put `SET TIME ZONE INTERVAL '+00:00' HOUR TO MINUTE` on the docs because it is a special statement we support for some client connectors. Putting it into the doc will make users misunderstand that RisingWave has supported these type of commands completely.

WanYixian:
> I think we don't need to put `SET TIME ZONE INTERVAL '+00:00' HOUR TO MINUTE` on the docs because it is a special statement we support for some client connectors. Putting it into the doc will make users misunderstand that RisingWave has supported these type of commands completely.

Makes sense, I'll keep it as it is

PR #2736, opened 2024-10-29
Doc: support subscription cursor to be blocking

### LINKED ISSUE #2672: Document: feat(subscription): support blocking cursor
This issue tracks the documentation update needed for the merged PR #18675.

Source PR URL: https://github.com/risingwavelabs/risingwave/pull/18675
Source PR Merged At: 2024-10-08T14:43:35Z

If it is a major improvement that deserves a new page or a new section in the documentation, please check if we should label it as an experiment feature.

WanYixian:
## Description

Divide `fetch from cursor` to be non-blocking and blocking

## Related code PR

https://github.com/risingwavelabs/risingwave/pull/18675

## Related doc issue

Resolve https://github.com/risingwavelabs/risingwave-docs/issues/2672

## Rendered preview

https://pr-2736.d2fbku9n2b6wde.amplifyapp.com/docs/next/subscription/#fetch-from-cursor

## Checklist

### DIFF
--- a/docs/transform/subscription.md
+++ b/docs/transform/subscription.md
@@ -88,37 +88,45 @@ FETCH from cursor function is currently only supported in the PSQL simple query 
 
 #### Non-blocking data fetching
 
-After creating a subscription cursor, you can fetch the data by the `FETCH NEXT FROM cursor_name` command. It returns one row of incremental data from the subscribed table. See the example below:
+After creating a subscription cursor, you can fetch the data using the `FETCH NEXT FROM cursor_name` or `FETCH n FROM cursor_name` command. These are non-blocking methods for retrieving data. Even if no new data is available, it returns an empty row immediately. When new incremental data is available, calling this statement again will return the latest row of data.
 
-```sql
-FETCH NEXT FROM cur;
+1. `FETCH NEXT FROM cursor_name` command returns one row of incremental data from the subscribed table.
 
-----RESULT
-t1.v1 | t1.v2 | t1.v3 |    t1.op     |  rw_timestamp  
--------+-------+-------+--------------+---------------
-     1 |     1 |     1 | UpdateDelete | 1715669376304
-(1 row)
-```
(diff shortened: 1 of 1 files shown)


WanYixian:
Hi @hzxa21 and @xxhZs , for [timeout not set](https://pr-2736.d2fbku9n2b6wde.amplifyapp.com/docs/next/subscription/#fetch-from-cursor) part below,
image
My reference is the release note of this [code pr](https://github.com/risingwavelabs/risingwave/pull/18675)
image
My question is, when timeout is not set, can it just be equivalent to `FETCH NEXT/n FROM cursor_name`, that is, the non-blocking way of data fetching? Thanks in advance

hzxa21:
Thanks for raising the question. I think what you have documented is correct but I do feel that it can confuse users with: if timeout not set in case 1, timeout = max; if timeout not set in case 2, timeout = 0. 

My suggestion is to clearly document the behavior of blocking fetch and non-block fetch, without mentioning "if timeout not set". I will put an example with my own words and feel free to modify it:
### Non-blocking data fetch
...
Fetch at most N rows from the cursor. If there are less than N rows available, return whatever we get immediately without waiting. This also means if there is 0 row available (already reached latest), return immediately with empty result.
...

### Blocking data fetch
...
Fetch at most N rows from the cursor with timeout. The fetch statement will be returned when either the N rows has been fetched or the timeout fires. When the timeout fires, whatever has been read so far will be returned. There are two cases when the timeout can be hit:
1. The cursor already reached latest and is waiting for new data to come for too long.
2. There are at least N rows available for cursor to read but reading all of them takes a long time.

You can use a large timeout to mimic the case when you would like the `FETCH` to block until new data arrives without needing to frequently "poll" subscription for new data with the non-blocking `FETCH`.
...

WanYixian:
Closed in https://github.com/risingwavelabs/risingwave-docs-new/pull/23

PR #2739, opened 2024-10-30
Update ingest-from-mysql-cdc.md, we don't support complex data types yet

### LINKED ISSUE #2738: Bug: Update MySQL schema mapping
### Where's the bug?

https://docs.risingwave.com/docs/current/ingest-from-mysql-cdc

### Describe the bug

Some datatypes are not actually supported, and would be polyfilled with `null` values.

https://risingwave-labs.slack.com/archives/C03A2PSS8KU/p1730208229551749

StrikeW:
complex data type are not supported

## Related doc issue

close https://github.com/risingwavelabs/risingwave-docs/issues/2738
Actually, docs of old release versions should also be changed.

## Checklist


WanYixian (approved):
Applied changes to older versions, thanks

